PR #4, opened 2021-12-02
Python 3 Dockerfile

ric-evans:
The Dockerfile is now Python 3 compatible.

I've also stripped away dependencies that were difficult to work with (or also not py3-compatible). Some of these may be added back in future revisions. For now, everything "compiles"

All changes are being made in the new `skymap_scanner/` directory. This is where all the skymap scanner py files will live after the two versions are remedied.


blaufuss (approved):
Looks good, pulling this in before I start bringing in things from SVN. Nevermind, it's still a draft...

ric-evans:
See updated description at top

Issue #6, opened 2022-01-30
Generate simple test case scan


blaufuss:
@ric-evans needs a simple test case for CI testing as he works on distributed workflows.  Ideally this scan would be simplified so that:
* Runs in ~10 minutes or so
* Uses ~10 client jobs
* Has a well defined set of inputs (i3/GCD files)
* Has an expected result that can be checked at completion

Suggestion discussed:
* Nside = 1 scan (12 pixels)
* Mark large fraction of DOMs as bad in GCD if more speedup is needed.

It's understood that first version will be condor-job based, but will allow for migration to docker/cloud based workflows

PR #11, opened 2022-01-31
Add simple test case scan

ric-evans:
closes https://github.com/icecube/skymap_scanner/issues/6


ric-evans:
@clagunas can you also add `pycache/` to the `.gitignore`?

ric-evans:
> @clagunas can you also add `pycache/` to the `.gitignore`?

Never mind, I'll make a branch with a full `.gitignore`

blaufuss:
@clagunas are these copies of scan scripts from other areas?  If so, I'd worry if these are updated/tweaked they fall out of sync with the "main scanner" scripts.  Is it not possible to call the production scripts with appropriate options and reduce duplication?

clagunas:
They are modified copies, yes. As it is right now, it's not possible. For some of the functionalities I would have to change the main scripts so that we can parse the options, but for others I don't know if I can. That's why in principle I wasn't planning to merge the two branches. Let me take a look at it this week and see what I can do. Anyway, the changes in those scripts are minimal.

ric-evans:
> They are modified copies, yes. As it is right now, it's not possible. For some of the functionalities I would have to change the main scripts so that we can parse the options, but for others I don't know if I can. That's why in principle I wasn't planning to merge the two branches. Let me take a look at it this week and see what I can do. Anyway, the changes in those scripts are minimal.

If the only differences between the testing scripts and the "main scanner" scripts (those in `resources/scripts/`) are the command-line arguments/options, it could be useful to add a `common_script_utils` folder to the "main scanner" scripts which hold the common classes/functions. Then, import and invoke those as appropriate. Thoughts?

dsschult:
It might also be useful to make a branch with the changes made directly, so we can see a proper diff of the changes.  Even if we never merge that, we could use that diff to see where to make the changes in the future.

Specifically, @ric-evans will likely be taking a chain saw to a large swath of the connecting code, and it might be easier to make the changes for these options during/after that.

clagunas:
I added comments that start with # test-case-scan so that they are easy to track down. The only file that I haven't commented yet is perform_scan because that's the one that changed the most and I'm still looking into it.

Issue #19, opened 2022-03-08
IceTray project or python package


mlincett:
Following up on issue #13 and the discussion on Slack, a decision should be made about whether maintaining the `skymap_scanner` as a python package or as an IceTray project. In the second case, it could be integrated into `realtime_tools`.

@dsschult suggested to split the project between a pure-python _server_ side an IceTray _client_ handling the actual LLH scan. This could simplify the implementation of CI testing.

I am opening this issue for further discussion.

ric-evans:
I strongly advocate separating `skymap_sanner` as its own python package. I've run into the scenario @dsschult mentioned about CI testing. It's much simpler adding icetray (`icecube`) as a resource, then building it with `skymap_scanner` included as a project.

ric-evans:
CI testing is our best bet at measuring how this all will function once it's cloud-bound / in other strange new places

ric-evans:
Also, `cloud_tools` did exactly this. They use `skymap_scanner`-type imports (technically `pulsar_icetray`), not `icecube.skymap_scanner` imports

kjmeagher:
I think making this a python package is a good idea as well. There seems to be a general trend of people preferring code which is independent of IceTray and this code seems to be only loosely coupled with IceTray

PR #24, opened 2022-04-06
Modularized Python-Package (Groundwork)

ric-evans:
- Introduce python package (`skymap_scanner`) with subpackages `skymap_scanner.server` and `skymap_scanner.client`
   + These are meant to be used together in a 1-N server-client model.
   + Both of these subpackages are individual services that can be started using, for example: `python -m skymap_scanner.server`
- Add CI actions for launching both services
- Expand `Dockerfile` along with `common_entrypoint.sh`, to startup either service with minimal redundancy

closes https://github.com/icecube/skymap_scanner/issues/19


ric-evans:
Since the SVN merge-over didn't preserve file/line history, I'm not considering that to be important as I copy files over.

ric-evans:
In an effort to keep PRs small(er), I'm moving the edited python files to a new PR and reverting these changes here

Issue #25, opened 2022-04-11
Config of GCD path(s)


mlincett:
Going through duplicate and divergent files (#5), here's a summary on how the path of GCD files is configured in `conf.py`:

`cloud_tools` uses a variable named `base_GCD_path`, in turns loaded in:
* `cloud_tools/calculate_online_alert_dict.py`
* `cloud_tools/extract_i3_file.py`
* `cloud_tools/prepare_frames.py`
* `cloud_tools/save_pixels.py`
* `cloud_tools/scan_pixel.py`

`skymap_scanner/python` uses a variable (list) named `GCD_base_dirs`, which is loaded in:
* `python/extract_json_message.py`
* `python/load_scan_state.py`
* `python/perform_scan.py`
* `python/utils.py` (`cloud_tools/utils.py`)
* `test-case-scan/extract_json_message_mod.py`
* `test-case-scan/perform_scan.py`

ric-evans:
Thank you @mlincett! This is very helpful

Issue #2, opened 2021-12-02
Implement MQClient In Place of Pulsar


ric-evans:
MQClient is an API for a message-queue abstraction. This will allow for cloud-based deployment, or local deployment, among a variety of MQ services (rabbitmq, pulsar, gcp, etc).

ric-evans:
We will use `wipac-mqclient-pulsar` and/or `wipac-mqclient-pulsar[telemetry]`

Issue #26, opened 2022-04-12
Divergent versions of `extract_json_message.py`


mlincett:
The file `extract_json_message.py` is divergent between `python` and `cloud_tools`, so I am trying to summarise the functionalities.

The file mainly provides the `extract_json_message()` method, that in turn recalls `full_event_followup.i3live_json_to_frame_packet()`.

The version in `cloud_tools`: 
- calls `full_event_followup.i3live_json_to_frame_packet()`
- checks that the 4th entry in the frame packet is a DAQ (Q) frame
- reverts the name of `QI3EventHeader` to `I3EventHeader` (if applicable)
- simply returns the frame packet.

The version in `python`:
- checks if the cache dir is valid
- calls `full_event_followup.i3live_json_to_frame_packet()`
- parses the state packet into a `event_id` and a `state_dict` with the function `__extract_frame_packet()`.
- returns `load_scan_state()` against `event_id` and `state_dict`.

Coming to the auxiliary function `__extract_frame_packet()`:
- checks that length of frame packet = 5
- checks that last frame in frame packet is a physics (P) frame
- _move the last packet frame from Physics to the Physics stream_ by calling `rewrite_frame_stop()` (??)
- creates `event_id_string` from physics frame header
- handles / creates the individual event cache under the general cache dir
- lot of complicate handling of GCD / GCD diff file (to be further investigated)
- calls `prepare_frames()` that apparently returns (modified?) the `frame_packet` and an `Excluded_DOMs` object.

Clearly the divergence reflects different requirements between the two modes (realtime and offline), and it seems there is room for better code organisation and documentation.

mlincett:
> One thing I feel is missing is a high level overview of what information is needed and what the message should contain for the scan to work.

If I understand correctly, in this context the _message_ is just a JSON dictionary carrying I3 frames in binary format (see [example](https://convey.icecube.wisc.edu/data/user/ckopper/event_HESE_2017-11-28.json)).

I am not aware of how this can be baked from an I3 file.

ric-evans:
Note: https://github.com/icecube/skymap_scanner/blob/master/skymap_scanner/extract_json_message.py was copied from https://github.com/icecube/skymap_scanner/blob/master/python/extract_json_message.py

mlincett:
Code has been largely reorganised so this is no longer an "issue".

PR #30, opened 2022-04-29
Communication Layers Refit

### LINKED ISSUE #5: Clean up repeated files from cloud_tools/ and python/
There are several files that were copied to cloud_tools from old skymap_scanner that appear to be old and/or modified copies of old scripts from ancient versions of skymap_scanner from SVN.  These need a more careful look as things move forward.  I suspect some pieces of these might be needed as things transition to cloud again with modern scan.  But "standard condor" scan will be from python/ and resources/ dirs right now (as they're currently run)

Please consider these repeated files also present in python/:
- cloud_tools/
    config.py
    extract_json_message.py
    perform_scan.py
    scan_pixel.py/scan_pixel_distributed.py (and python/traysegments/scan_pixel_distributed.py)
    utils.py

Several commits from SVN history might be helpful to see what's changed in the Condor scanner since it split:
https://code.icecube.wisc.edu/projects/icecube/log/IceCube/sandbox/ckopper/skymap_scanner?rev=185232

### LINKED ISSUE #13: Clean up of symlinks
Several files in `resources/scripts` are symlinks to `python`. This likely requires a cleanup.

### LINKED ISSUE #27: Divergent versions of `perform_scan.py`
The `perform_scan.py` differs between `cloud_tools` and `python` (see #5).

The `cloud_tools` version is designed to operate with a fixed NSIDE value, while the `python` version has the machinery to perform the hierarchical scan as we do in real time.

The core functionality is provided as an IceTray module named `SendPixelsToScan`.

The input is prepared by `extract_json_message()` (see #26). For `cloud_tools`, this is simply the GCDQp frame packet. For `python`, the `state_dict` is loaded.

The version in `cloud_tools` has added reco functionality (see `get_reco_losses_inside()`). This part could be easily merged.

ric-evans:
Implement MQClient and add simplified GCD communication/transfer

The main files are:
- `skymap_scanner/server/start_scan.py`
	+ takes a realtime pkl event, sends pixels to clients, and saves the best scan for each
	+ think of this as a de-icetray'd `perform_scan.py`
- `skymap_scanner/client/client.py`
	+ the end-point of the pulsar communication
	+ spawns a subprocess for a millipede tray for processing a single pixel (variation)
- `skymap_scanner/client/scan_pixel_pkl.py`
	+ the millipede tray
	+ essentially unchanged from `scan_pixel_distributed.py` (updates from the slack channel)

@mlincett & @dschull: please look at "TODO" comments, most of these are things I haven't figured out

closes #27
closes https://github.com/icecube/skymap_scanner/issues/25
closes #13 (no longer needed)
closes #5 (no longer needed)
closes https://github.com/icecube/skymap_scanner/issues/2


dsschult:
Is it a good idea to try putting MQClient as IceTray modules?  Or is there a better way to do this outside IceTray?

ric-evans:
> Is it a good idea to try putting MQClient as IceTray modules? Or is there a better way to do this outside IceTray?

This was my plan for the client-side. Is the alternative something like this:

ric-evans:
@dsschult I'm most unsure about how to get the frame/scan out of the tray. But also, I don't know if there's any significant performance cost to building & tearing down icetray for each pixel/incoming-message.

kjmeagher:
There isn't much overhead for instantiating a new tray for every message, but instantiating a lot of trays has been known to cause memory leaks in the past. I am not aware of any problem with having one icetray module listen for messages. and having another module that sends messages after the reconstruction

dsschult:
One other alternative is just taking the hit to starting a subprocess for each msg in the stream.  That's known safe, and if libraries get cached in memory is still fairly fast.

An example for the client:
```python
for in_msg in stream:
  with open('in_msg','w') as f:
    f.write(in_msg)
  subprocess.check_call(['icetray_client.py', 'in_msg', 'out_msg'])
  if os.path.exists('out_msg'):
    with open('out_msg') as f:
      out_msg = f.read()
    back_to_client.send(out_msg)
```

ric-evans:
Note: the `Signals.SIGBUS: 7` error may be due to a serialization bug. Currently, the I3Frame is naively pickled and sent over the wire, the frame once deserialized may have dangling/invalid C++ pointers. We'll need to not send I3Frames, but rather something smaller that can be _made into_ an I3Frame. 

@briedel brought up a similar point in https://github.com/icecube/skymap_scanner/issues/26#issuecomment-1096756587

ric-evans:
CI is working and will be sped up. I am planning to add more components to make the jobs more "test" like

dsschult (commented):
I reviewed the three main files, and it looks good to me.

ric-evans:
@mlincett I've largely preserved the reporting and logging callbacks, see `ProgressReporter`. We will need to decide on a design for how to "pass in" the callback functions. This will ultimately depend on how we execute the scanner. The scanner is started as an importable module in the (pre-existing) alert listener (2.0). I don't think this is scalable. Instead, I'd prefer to fire off a subprocess script that will start up the server and n clients. This way if the alert listener goes down, the server is still operational.

Issue #9, opened 2022-01-30
Scanner output formats need to be improved/expanded


blaufuss:
The current i3 file output from scans is difficult to interpret by non-experts, and ideally we need to be
able to automate "after the scan" workflows which are highly manual right now.

Suggestions here include:
* Output of a FITS skymap with key information in the header (ala catalog files in /data/ana/realtime/alert_catalog_v2/fits_files)

Potential after scan automation needs include:
* consumption by an alert DB
* consumption by I3Live
* generation of followup GCN/TACH notices

blaufuss:
Another potential consumer is SCIMMA HopSkotch

mlincett:
Note (mostly to myself) that some code is available from the [v2 alert catalog production scripts](https://github.com/icecube/alerts_processing).

ric-evans:
@mlincett can we close this now? We have the `.npz` output file

ric-evans:
We now also have `ScanResult.read_json()` and `ScanResult.to_json()` for communication with SkyDriver (#109)

Also, `scan_result.py` could be moved to a new `skymap_scanner_tools` repo for portability.

I'm closing this issue since output formats are A. much better now and B. further improvements are outside the scope of the scanner proper

PR #40, opened 2022-06-19
Scan result to single numerical file

mlincett:
Currently, all scripts that use the scan result (plotting, finding best LLH, etc.) parse the cache directory of an event (few hundreds of MiBs in size) to obtain the required numerical values (notably the `llh` is used to determine the fit quality for the position of the pixel).

This PR introduces a `ScanResult` class that can be instantiated by reading an event from a cache directory (see associated `save_scan_result.py`) and saved to a `numpy` file (`.npz`) holding a dictionary of structured arrays (one per `nside`).

The same class can be instantiated by loading the `numpy` file itself and two `ScanResult` instances can be compared with either `__eq__` / `==` or `is_close()` methods.

In other words, **we make it easy and efficient to save and compare scan results** without having to load a huge cache directory. In the future, the class can be augmented with plotting / FITS output / whatever we find useful.

(Note that the equality fails whenever a `NaN` is present in the result, because `numpy.array_equal()` does not allow for NaN tolerance before 1.19 (current build has 1.17.4). `is_close()` will however work.)


mlincett:
@ric-evans I wanted to put `save_scan_result.py` in a dedicated `scripts` directory, however it seems that only modules can use relative import paths. In other words, `from ..scan_result import ScanResult` would not work. For this reason I had to put `save_scan_result.py` and `scan_result.py` in the same directory (under `skymap_scanner` together with the other common files).

I am not sure how to deal with this inconvenience, maybe we should create a `setup.py` so the `skymap_scanner` package can be installed in development mode?

skymap_scanner/save_scan_result.py
```diff
@@ -0,0 +1,55 @@
...
+    close = result.is_close(result_check)
+    equal = result == result_check
+
+    logger.info(
+        f"The loaded file is close? ({close}) and equal? ({equal}) to the source data."
+    )
+
+
+if __name__ == "__main__":
+    main()
```
ric-evans, line 55: Instead of creating a `.npz` file by loading from the `state_cache` after each run, how about we use `ScanResult` as a final step in the main pipeline? Then we compare the two `.npz` files in a dedicated test script (call it `tests/compare_scans_npz.py`).

ric-evans (changes requested):
I'd prefer to not rely on `state_cache` after the pipeline has finished, where possible

ric-evans:
> Ideally it should serve as the basic data structure for plotting / processing / transmission of the scan result.

ric-evans:
> Ideally it should serve as the basic data structure for plotting / processing / transmission of the scan result.

I do agree with this 100%, minimizing the number of necessary outputted data files is great for loose coupling

ric-evans:
@mlincett what're your thoughts on https://github.com/icecube/skymap_scanner/issues/9?

mlincett:
> @mlincett what're your thoughts on #9?

The current plotting functionality and the FITS output should be integrated in `ScanResult`, but this will come later. I suggest we leave that open for now.

mlincett:
> I'd prefer to not rely on `state_cache` after the pipeline has finished, where possible

Indeed. As we discussed, `ScanResult` should be instantiated in the pipeline and saved to a `npz` file. The provided script is meant to read the caches of previous scans (now stored in the followup machine). A new script is now included to compare two `npz` files.

The return values of this script can be changed if needed.

ric-evans (approved):

ric-evans (approved):

ric-evans:
> When filling reco losses with zero values the test are successful. However, I would strongly advise we implement the calculation/storage of reco losses (#45) before moving on with testing the condor scan.

We'd still be able to get what we need if we persist the cache directory, correct?

ric-evans:
Let's wait to merge this branch, so we can implement #45 here. Sound good, @mlincett?

ric-evans:
closes #45

mlincett:
> Let's wait to merge this branch, so we can implement #45 here. Sound good, @mlincett?

Alright. #45 is taken care of and now passing the CI test. I'd say all looks good.

Issue #34, opened 2022-06-09
Add CI Tests (Compare Outputs)


ric-evans:
Let's add tests that will compare outputs to known outputs, for a mini-scale run.

Steps:
1. Run full-scale on HTCondor cluster (#35) and compare/confirm results (a la #40) to known results from `python/`-version
	- **UPDATE:** _this has turned out to be more complicated than it seemed, see #60_
3. Run mini-scale on CI (already implemented) and persist outputs
4. Add CI Test(s): Compare persisted mini-scale outputs against CI

When this is complete we can remove `test-case-scan/`, since all its code will be integrated into `skymap_scanner/` and `tests/`.

ric-evans:
Also, don't run the CI on "tags":

ric-evans:
waiting on https://github.com/icecube/skymap_scanner/pull/40

ric-evans:
waiting for https://github.com/icecube/skymap_scanner/pull/46

Issue #14, opened 2022-03-07
Configuration file


mlincett:
Some parameters that are currently hardcoded should be probably moved to a configuration file. Do we agree to this?

What is the fashionable config file format nowadays? YAML, TOML or other?

dsschult:
Are they constants, or config parameters like slack tokens? I'm guessing the latter.

Env variables are very popular with containers, but if you absolutely know you won't be using a container, then any of those config formats are fine.  Generally it depends on how complex your config is, and if it's only a few variables you can get by with the [Python configparser](https://docs.python.org/3/library/configparser.html).

I've also seen projects that will take both a config file and env variables, if you want to be fancy.

mlincett:
I will try to collect some info and see what could be better tracked in a conf while not being covered by current parser arguments.

The `config.py` route is quite straightforward but in general I prefer to keep open the possibility of having multiple presets. However, no strong opinion on this from my side.

mlincett:
A `config_listener.py` has been added for variables related to `v2_alert_listener.py` with [fa0ef39](https://github.com/icecube/skymap_scanner/pull/18/commits/fa0ef39154733ac1d327c87e8789f911c4f70a87).

Note that another (partially deprecated) `config.py` file exists under `python`, but in order to make effective any modification to its content one has to rebuild the metaproject (from which the config is actually imported).

mlincett:
Things that (may) need to be configured:

- number of nodes to use (?)
- slack ids of the shifters to be alerted with direct mentions (critical warnings only)
- `gfu_prescale`: a subset of all GFU events (one every n-th with `n=gfu_prescale`) can be used as a heartbeat. Such events should be pre-processed (TBD to what extent) but not scanned.

ric-evans:
> * number of nodes to use (?)

We haven't decided to make this a fixed number (even within a single event scan) ...I think that's a cool part of the MQ design :smile: it can grow dynamically and the alert listener wouldn't need to know how many there are, anyways this is going out of scope...

mlincett:
I think this issue has long been superseded. Closing.

PR #55, opened 2022-07-11
Add CI Tests for Outputs (`.npz`)

ric-evans:
closes https://github.com/icecube/skymap_scanner/issues/34

adds environment variables (see https://github.com/icecube/skymap_scanner/issues/14)


ric-evans:
UMD's power outage has killed docker builds for the time being (svn). Expected outage "until late Thursday".

ric-evans:
NOTE: since I changed the MQ attributes, we'll have to regenerate the `.pkl` files with _this PR's_ docker image (that is, when UMD has power)

mlincett (approved):

PR #59, opened 2022-07-15
Performance & Configuration Upgrades

### LINKED ISSUE #50: Queue Naming Scheme
We need a naming scheme for queues. We can build a name from the "event id", "nside min", "nside max", etc. There's a 20-char limit for queue names, so a hash of a few variables may be best (event ids are long). This "built name" is the _basename_, we'll always prefix `to-clients-`/`from-clients-`.

One Design Option:
1. server starts up
2. server creates queue basename, and writes to `skymap-scanner-queue-basename.txt`
3. now clients can be started using by `cat skymap-scanner-queue-basename.txt`

### LINKED ISSUE #56: Update README
The README hasn't been updated for v3. We can also use https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/ to show a flowchart of the processes

### LINKED ISSUE #57: Decrease Message Size
> In testing we've used two types of messages (incoming and outgoing). The first ranges ~100kB to ~700kB. The second, ~20kB-120kB. [...] we may be able to get the messages down to ~100kB and ~1kB, respectively.

#### Current Design
- to-client message (1-per-pixel) = geometry frames + gcd strings + the pixel's frame
   * ~100kB to ~700kB
- to-server message (1-per-pixel) = whole pixel's frame w/ added reco info
   * ~20kB-120kB

#### Proposed Design
- "startup" data (1-per-client) = geometry frames + gcd strings
   * ~100kB to ~700kB
   * a pkl file provided via command-line args
   		- we'll need to change the start-up pattern so we can grab the pkl from the active server (similar to #50)
- to-client message (1-per-pixel) = the pixel's frame
   * ~100kB
- to-server message (1-per-pixel) = a few ids , llh, recoLossesInside, and recoLossesTotal
   * ~1kB
   * this means no longer saving i3 files for each pixel
   * need @mlincett's input (do we need to be saving i3 files?)

### LINKED ISSUE #58: Add GCD Files to Docker Container
It's ideal if the worker nodes (clients) do not need to be mounted to CVMFS.

The GCD files in `PoleBaseGCDs/` are about ~306MB. The image is ~6GB. Let's throw it in there.

A future solution could involve using a remote object server (blob, s3 bucket, ...). This would remove the need for adding the spline tables to the container.

ric-evans:
#### Programmatically Name In/Out Queues
- closes #50 

#### Decrease Message Size
- closes #57 
- no longer sending full reco frames to the server, or saving _n_ `.i3` files to the cache dir
- changed the "nside dict" type from `Dict[int, Dict[int, I3Frame]]` to `Dict[int, Dict[int, PixelReco]]` (@mlincett)

#### Add GCD Files to Docker Container
- closes #58

#### Alternate Reconstruction Algorithm Support
- see #8 
- @mlincett, this will need further investigation to see if I siloed millipede-specific logic completely

closes #56


ric-evans:
I've added many constants to `config.py` (https://github.com/icecube/skymap_scanner/issues/14)

ric-evans:
@mlincett I've removed many uses where the `state_dict` is being passed to a function when the nside-dict (or other members) could be passed instead (mainly `server/` modules)

ric-evans:
@mlincett I've also removed the `"MCradec"` key from the state_dict as we discussed. We'd need to pass RA and Dec values onto `choose_new_pixels_to_scan()` if/when we want to support this again (like, giving the floats to `server` via a new command-line arg)

mlincett (approved):
Nice work! At a first glance I see no issue with the PR.

mlincett:
About #8 there is some millipede-specific login in `skymap_scanner/utils/prepare_frames.py`. Possibily different reconstruction algorithms could need to trigger different preprocessings of the frames. Maybe we can achieve that by defining IceTray segments that can be selected. I had started modularising this in the [listener_overhaul](https://github.com/icecube/skymap_scanner/tree/listener_overhaul) branch but I didn't have much time to work on this lately.

ric-evans:
> About #8 there is some millipede-specific login in `skymap_scanner/utils/prepare_frames.py`. Possibily different reconstruction algorithms could need to trigger different preprocessings of the frames. Maybe we can achieve that by defining IceTray segments that can be selected. I had started modularising this in the [listener_overhaul](https://github.com/icecube/skymap_scanner/tree/listener_overhaul?rgh-link-date=2022-08-12T09%3A57%3A06Z) branch but I didn't have much time to work on this lately.

Okay, I'll leave this for you in a future PR. I agree that programmatically selecting different IceTray segments is the way to go, see https://github.com/icecube/skymap_scanner/blob/65441126f1bc7a8ed18ab4044ce99127527a037f/skymap_scanner/client/reco_pixel_pkl.py#L207

PR #63, opened 2022-08-19
Remove Legacy Files

### LINKED ISSUE #36: Remove Deprecated Code
Any files that need to persist should be moved to `skymap_scanner/` or its own directory with a descriptive name. This is also a good time to remove any dead legacy code. One option is to run multiple test runs with a coverage tool.

ric-evans:
Any utility-type scripts we want to keep can be moved to `scripts/utils/`. Those files will also need to be updated to match the v3 design, e.g. the new `nsides_dict` (`NSidesDict`/`PixelReco`) structure.

closes #36


mlincett:
I think we should remove `python` only after the new version is in production.

ric-evans:
> I think we should remove `/python` only after the new version is in production.

That's a good point. Let's keep them for a little while, but clearly indicated this is old code. https://github.com/icecube/skymap_scanner/pull/70

mlincett:
I provided at [skymap-scanner-legacy](https://github.com/icecube/skymap-scanner-legacy) a minimal interface to the SVN version in CVMFS, and produced a tag `v1.5` that can be used to track back the old codes. I think there is no more reason to keep `deprecated_files` in the main branch.

PR #68, opened 2022-10-10
Production Readiness

### LINKED ISSUE #48: Add Singularity Compatibility
The grid (OSG) is Singularity-friendly, not docker. Look into either converting docker images (preferred but slower?) or using a native Singularity solution.

ric-evans:
Systems-level things for a production environment:
- Singularity Ccompatibility (closes #48)
- Updated Condor Scripts
- Documentation

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,13 +20,16 @@ env:
 
 jobs:
 
-  docker-build:
+  build-containers:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - name: Build Docker Image
         run: |
           docker build --no-cache -t icecube/skymap_scanner .
+      - name: Build Singularity Container
+        run: |
+          sudo singularity build skymap_scanner.sif docker-daemon://icecube/skymap_scanner
 
   test-run-all:
     runs-on: ubuntu-latest


ric-evans:
TODO: Use an object-based design for different recos. Consolidate all the `if reco_algo == cfg.RecoAlgo.*` logic into class/derived-classes.

ric-evans:
**TODO:** Update condor script(s)

ric-evans:
**TODO:** Update docs

.github/workflows/tests.yml
```diff
@@ -126,16 +199,37 @@ jobs:
...
+          cp tests/data/reco_pixel_pkls/get_toclient_msg_pkl.py \
+            tests/data/reco_pixel_pkls/${{ matrix.dir }}
+
+          # create the in.pkl
+          docker run --network="host" --rm -i \
+            --shm-size=6gb \
+            --mount type=bind,source=$(readlink -f tests/data/reco_pixel_pkls/${{ matrix.dir }}),target=/local/pkls \
+            --env PY_COLORS=1 \
+            $(env | grep '^SKYSCAN_' | awk '$0="--env "$0') \
+            icecube/skymap_scanner:latest \
```
dsschult, line 217: It seems wrong that we'd use a docker image from `main`.  What if the PR is editing something that affects this?

Is there actually a reason we need to use docker?  Would a local python install or venv not suffice?

skymap_scanner/client/client.py
```diff
@@ -41,7 +63,9 @@ def _create_dir(val: str) -> Optional[Path]:
             "The directory with the 'startup.json' file to startup the client "
             "(has keys 'mq_basename', 'baseline_GCD_file', and 'GCDQp_packet')"
         ),
-        type=Path,
+        type=lambda x: wait_for_file(
+            x, cfg.env.SKYSCAN_CLIENT_WAIT_FOR_STARTUP_JSON_SEC, "startup.json"
+        ),
```
dsschult, line 68: What is the purpose of this wait?  I can't imagine this is useful in production, since the startup file needs to be transferred with the job.

ric-evans: If we put the startup.json somewhere else, like `/data/user/`, then this is useful. It also means we can launch everything at once, so the clients can startup while the initial pixels are being generated (which can take a while on a full-size scan)

.github/workflows/tests.yml
ric-evans: It needs `icecube`/`icetray`

ric-evans: In contrast, `tests/compare_scan_results.py` doesn't need `icecube` (see several lines above)

skymap_scanner/client/client.py
ric-evans: yeah, I see your point now. Moving that logic upstream...

.github/workflows/tests.yml
ric-evans: > It seems wrong that we'd use a docker image from main. What if the PR is editing something that affects this?

Can you elaborate on this?

dsschult: Ah, one thing I think I didn't catch. You're running 
```
docker build --no-cache -t icecube/skymap_scanner .
```
in a previous step, which locally builds `icecube/skymap_scanner:latest`.

I'd recommend using a different tag, to disambiguate from the global `icecube/skymap_scanner:latest` that gets built for the `main`/`master` branch and pushed to dockerhub.  Because you technically can't guarantee with this usage that it's not pulling from dockerhub.

scripts/condor/spawn_condor_clients.py
```diff
@@ -49,23 +46,13 @@ def make_condor_file(  # pylint: disable=R0913,R0914
...
-            f"--auth-token {auth_token} "
-            f"--log {log} "
-            f"--log-third-party {log_third_party} "
-            f"--timeout-to-clients {timeout_to_clients} "
-            f"--timeout-from-clients {timeout_from_clients}"
-        )
         file.write(
-            f"""executable = {executable}
-arguments = {args}
+            f"""executable = python -m skymap_scanner.client
```
dsschult, line 53: The `executable` can be only one thing, so `python` in this case.  You'd need to add the rest to `arguments`.

dsschult (changes requested):
One question re the condor jobs: is there a reason not to use the condor python bindings to submit the jobs?  I guess it's nice to see the submit file, but not sure it's required.

scripts/condor/spawn_condor_clients.py
ric-evans: Okay :+1:

.github/workflows/tests.yml
ric-evans: Fair point, I suppose that's good practice in general not to tag a local image as `latest`

dsschult (approved):

Issue #8, opened 2022-01-30
Investigate migration of skymap_scanner to new reconsrtuction algorithsm under the hood,


blaufuss:
There are many reasons to think that the current millipede scan isn't perfect. We should investigate options
for:
* migrating to @nwhitehorn's new double precision millipede implementation in icetray/main
* SplineMPE scan might be a better idea.

Work with experts in the reco/systematics working group for suggestions and input.

mlincett:
- Re: migrating to millipede double precision, we can discuss this in the upcoming call/workshop. However, validation is an issue so we should agree on a sample of events that is sustainable to rescan in order to check agreement / improvement.
- Re: where to go in the future, I expect the discussion will take a while, however we can start restructuring the code so to make it modular / configurable at runtime, rather than having fixed hardcoded setups.

briedel:
Couple thoughts here:

Short-term: Change to splines/tables @tianluyuan suggested
Medium-term: See if we can get a track model of @mhuen cascade generator reco, so we have a fast reco with poorly constraint error that we can use to constrain the scan
Long-term: Somehow get a generator model to replace millipede

dsschult:
> * Re: where to go in the future, I expect the discussion will take a while, however we can start restructuring the code so to make it modular / configurable at runtime, rather than having fixed hardcoded setups.

Note that part of Ric's queue work will likely separate out the "client" part of the scan into an independent script.  That should make it easy to swap between algorithms.

ric-evans:
> > * Re: where to go in the future, I expect the discussion will take a while, however we can start restructuring the code so to make it modular / configurable at runtime, rather than having fixed hardcoded setups.
> 
> Note that part of Ric's queue work will likely separate out the "client" part of the scan into an independent script. That should make it easy to swap between algorithms.

Coming back to this: yes, this is relatively simple. 

On the app-server end, the only apparent assumption needing to change is relying on the `"MillipedeStarting2ndPass_millipedellh"` key for the llh value. On the client-side, we would swap the tray segments/modules for a non-millipede reco--easy. We'd also need to signal which reco we'd want to use (eg: add another attribute to the MQ payload)

ric-evans:
From a Slack thread with @mjlarson:

ric-evans:
[...]

PR #69, opened 2022-10-13
Update millipede

tianluyuan:
Updated settings for millipede, which include

- updated ice model spline
- refine seed particle time given direction and vertex position
- if positional variations exist, rotate and set in transverse plane
- minimizer initial step sizes for fixed direction
- simplex+migrad step
- lower bright DOMs threshold
- `RelUncertainty` setting

Branched from and merged after #68.


ric-evans (commented):
Is the `millipede2` traysegment the one currently in `master`?

skymap_scanner/server/start_scan.py
```diff
@@ -744,7 +808,7 @@ async def serve_scan_iteration(
                 except DuplicatePixelRecoException as e:
                     logging.error(e)
                 # if we've got all the pixrecos, no need to wait for queue's timeout
-                if len(col.pixreco_ids_received) == n_pixreco:
+                if len(col.pixreco_ids_sent-col.pixreco_ids_received) == 0:
```
ric-evans, line 811: you can use truthy set-logic here: `if A - B:`

skymap_scanner/server/start_scan.py
```diff
@@ -549,7 +601,12 @@ def collect(self, pixreco: pixelreco.PixelReco) -> None:
             raise DuplicatePixelRecoException(
                 f"Pixel-reco has already been received: {pixreco.id_tuple}"
             )
-        self.pixreco_ids_received.append(pixreco.id_tuple)
+        if pixreco.id_tuple not in self.pixreco_ids_sent:
+            raise DuplicatePixelRecoException(
+                f"Pixel-reco received not in sent set, it is probably from an earlier iteration: {pixreco.id_tuple}"
+            )
+            
+        self.pixreco_ids_received.add(pixreco.id_tuple)
```
ric-evans, line 609: I'm curious, were you seeing this scenario in a test run? I can see how it's possible since `MQClient` doesn't guarantee deliver-once delivery.

Also, see my comment below on `n_pixreco`/`pixreco_ids_sent`

tianluyuan:
‘millipede2’ is the new version and not on master yet

skymap_scanner/server/start_scan.py
```diff
@@ -695,17 +755,20 @@ async def serve_scan_iteration(
...
-                f"Sending message M#{i} ({pixelreco.pixel_to_tuple(pframe)})..."
+                f"Sending message M#{i} ({_tup})..."
             )
             await pub.send(
                 {
                     cfg.MSG_KEY_RECO_ALGO: reco_algo,
                     cfg.MSG_KEY_PFRAME: pframe,
                 }
             )
+            pixreco_ids_sent.add(_tup)
```
ric-evans, line 771: Following my comment above -- with `pixreco_ids_sent`, I'm not sure there's a need to have `n_pixreco`. We can replace all mentions of  `n_pixreco` with  `pixreco_ids_sent`, essentially all the way until the `ProgressReporter` instantiation.

ric-evans (changes requested):
Your work is looking great. And finally corralling all the old millipede logic is such a relief! A few comments on the MQ/`server` side of things

skymap_scanner/server/start_scan.py
ric-evans: or even `if A == B`, yeah?

ric-evans:
I think you've closed this issue, yeah? https://github.com/icecube/skymap_scanner/issues/8

skymap_scanner/server/start_scan.py
tianluyuan: implemented `A==B`

skymap_scanner/server/start_scan.py
tianluyuan: I removed the argument from the `PixelRecoCollector`, which I think is what you were suggesting?

skymap_scanner/server/start_scan.py
tianluyuan: Yes, I was seeing the server receive nside=8 pixels during the nside=64 iteration. The line above would only catch it if it was already in the list of received pixels. For the first (8, x, y) occurrence it would be added to the list, then later would raise a Duplicate exception when checking the `nsides_dict`. However, the length of the list was still incremented by 1.

tianluyuan:
I think #8 should be mostly solved yes. Reco specific logic can be defined in each individual file and by passing off the `reco_algo` to the `PixelsToReco` class it should be possible to set the seed and variation dependent on the reco_algo as well. `prepare_frames` also takes in `reco_algo` as well, that other recos could use.

skymap_scanner/server/start_scan.py
ric-evans: Looks great!

skymap_scanner/server/start_scan.py
ric-evans: I imagine this isn't happening _too_ often (thinking about performance). It's good we're accounting for it now.

ric-evans:
> I think #8 should be mostly solved yes. Reco specific logic can be defined in each individual file and by passing off the `reco_algo` to the `PixelsToReco` class it should be possible to set the seed and variation dependent on the reco_algo as well. `prepare_frames` also takes in `reco_algo` as well, that other recos could use.

closes #8

skymap_scanner/recos/__init__.py
```diff
@@ -51,7 +51,7 @@ def get_reco_interface_object(name: str) -> RecoInterface:
     """Dynamically import the reco sub-module's class."""
     try:
         module = importlib.import_module(f"{__name__}.{name.lower()}")
-        return getattr(module, name.capitalize())
+        return getattr(module, ''.join(x.capitalize() for x in name.split('_')))
```
ric-evans, line 54: nice! :sunglasses:

skymap_scanner/server/start_scan.py
```diff
@@ -448,7 +448,7 @@ def _gen_pixel_variations(
 
         for i in range(0,len(self.pos_variations)):
             posVariation = self.pos_variations[i]
-            if self.reco_algo != 'millipede':
+            if self.reco_algo != 'millipede_original':
```
ric-evans, line 451: My intuition is this could cause a future pain point. Can you make this explicit, i.e. use `==` only?

ric-evans (changes requested):

skymap_scanner/server/start_scan.py
ric-evans: Another option id to move this logic to the `RecoInterface`-based classes

tianluyuan: Good catch. I changed it to `==` for now.

ric-evans (approved):
Looks good. Feel free to squash merge when the last test passes. The singularity tests will need to be fixed in a future PR.

PR #82, opened 2022-11-07
Plotting functions in ScanResult

tianluyuan:
Port of the old plotting routines into the `ScanResult` class so that one can call
```
res=scan_result.ScanResult.load('run00000070.evt000000000000.HESE_8_64_512.npz')
res.create_plot_zoomed(dosave=True, plot_bounding_box=True)
```

Two methods are added, the `create_plot` method is the older standard matplotlib style plotting while the `create_plot_zoomed` uses healpy to produce a cartview and smoother contours with `meander`.

One thing that is missing is the MJD, as that no longer exists in `nside_dict`. Perhaps it's easiest to save it as part of the output filename of the result?


mlincett:
I am against encoding too much information in the filename (if any at all). We encoded the `nside` list mostly to avoid duplicating filenames when scanning the same event with different `nside` sequences.

I see two options:
1. extend `ScanResult` in order to incorporate the event metadata (it may be tricky to append them to a `npz` file, though);
2. maintain `ScanResult` as simple as possible, and (optionally) pass the event metadata to the relevant plotting methods.

I would prefer to go with the second option.

tianluyuan:
Yea I agree encoding metadata in filenames is suboptimal. Apparently, numpy arrays can store [metadata](https://numpy.org/doc/stable/reference/generated/numpy.dtype.metadata.html#numpy.dtype.metadata), so I think this could be something to try as I think it makes sense to have some association between the nside-X arrays and the event-id, mjd etc.

tianluyuan:
I implemented a version with the array metadata in #88. Please lmk if that looks ok to merge and I can update this PR accordingly.

tianluyuan:
I was able to run a scan using the latest build on cvmfs and create a plot using metadata that now contains the MJD. I think this should be good to merge?
Screen Shot 2022-11-09 at 22 09 57

mlincett:
Looks great. Feel free to merge!

Issue #7, opened 2022-01-30
skymap_scanner should take simulation or data as input without issue


blaufuss:
Not 100% sure this is not already true, but we should make sure this ability is not lost.

ric-evans:
@tianluyuan is #83 sufficient to close this issue? If so, is there a way to detect whether an event is simulated? See https://github.com/icecube/skymap_scanner/blob/f1f10fda0e00994b08a7cc2ff3a2c335070968d4/skymap_scanner/utils/extract_json_message.py#L130

tianluyuan:
Yes, I have been processing simulated events so to my mind this can be closed, or made more specific as to what might be missing still.

I'm not sure if there's a 100% full-proof way to detect if an event is simulated or not, since frame objects can be deleted and run/event ids can collide. The easiest might be to just check if the run-id falls within the data runs. I think that's another issue though.

mlincett:
Since there is a lot of variability in how the frame structure can look like, I would avoid trying to be "too smart" here and just require the user to flag if an event is simulated.

Issue #32, opened 2022-06-03
Scan preparation: event dictionary dumped to JSON and back to dictionary again


mlincett:
The `extract_json_message()` function that is run in the preparation of an event for the scan calls the following function:

`frame_packet = full_event_followup.i3live_json_to_frame_packet(json.dumps(json_data), pnf_framing=True)`

1. what is the virtue of converting to JSON the event we received from ZMQ?
2. is this operation supposed to mutate the data format / frame arrangement?
3. where is the JSON format used otherwise?

mlincett:
Giving a look at the source of [i3live_json_to_frame_packet](https://github.com/icecube/icetray/blob/main/full_event_followup/python/frame_packet_to_i3live_json.py), the only non-trivial operation that is performed is the `zlib` decompression.

The rest of the code is fairly redundant with what the skymap scanner code already does to check the integrity of the packet.

ric-evans:
realtime's json format is a standard, so let's preserve it despite its trivialness. I agree the `json.dumps()` + `json.loads()` is inefficient, but I'd argue [i3live_json_to_frame_packet()](https://github.com/icecube/icetray/blob/main/full_event_followup/python/frame_packet_to_i3live_json.py#L88) should be able to take an object as input.

I'm closing until we see empirical evidence that this is something that needs to be optimized.

PR #83, opened 2022-11-07
Add an i3_to_json converter script

tianluyuan:
This is a rough script to read in i3 files and convert it into realtime's json format for use with the skymap scanner. Perhaps it could be pieced apart into the scanner itself at some point.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -36,7 +36,7 @@ iminuit==2.17.0
     # via skymap-scanner (setup.py)
 kiwisolver==1.4.4
     # via matplotlib
-matplotlib==3.6.1
+matplotlib==3.6.2
     # via healpy
 numpy==1.23.4
     # via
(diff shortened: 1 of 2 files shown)


ric-evans:
Could you make this a module in `skymap_scanner/utils`? This way we can give the scanner a `.i3` directly.

ric-evans:
This has been requested several times, so let's fast-track it.

It will close https://github.com/icecube/skymap_scanner/issues/7, right @mlincett @tianluyuan?

ric-evans:
@mlincett's comments in https://github.com/icecube/skymap_scanner/issues/32 are also relevant here

skymap_scanner/server/start_scan.py
```diff
@@ -1013,18 +1013,30 @@ def _validate_arg(val: T, test: bool, exc: Exception) -> T:
         raise NotADirectoryError(args.gcd_dir)
 
     # read event file
-    if args.event_file.suffix == ".json":
-        # json
+    # .I3
+    if args.event_file.suffix == ".i3":
```
tianluyuan, line 1017: We should probably allow the suffix to be ".i3" or ".i3.[gz/zst/bz2]".

ric-evans (approved):
Let's integrate this into the Skymap Scanner Service when it arrives. For now a script is good

PR #88, opened 2022-11-07
Preserve useful metadata in the nside array

tianluyuan:
Numpy arrays can store metadata, though not preserved under all operations. It is sufficient, however, for storing event information and other useful metadata inside ScanResult's nside array.


mlincett (approved):
Good, I think this is a viable solution for the time being.

PR #89, opened 2022-11-08
Alternative test data

mlincett:
This PR adds a second scan result for the two benchmark events.

Replaces #63 (too outdated to fix).


ric-evans:
@mlincett is this still wanted?

mlincett:
> @mlincett is this still wanted?

I guess not. We should rather aim to understand the variability between result and keep a single sample of test data for each reco.

PR #98, opened 2022-11-29
Fix ci

dsschult:
Switch around the plotting imports so they are local to those functions.   This allows the file to be imported without icetray, so results can be verified.


dsschult:
Two checks failed for running out of space.  Changing the docker image should fix this.

PR #99, opened 2022-11-30
Update a few settings to optimize contours and speed

### LINKED ISSUE #86: Try double precision millipede
In tests outside of scanner bot, double precision millipede seems to help with minimizer convergence. This would be nice to have for test optimization.

tianluyuan:
Updates to use double precision photospline and tighter tolerances. Simplex precision unchanged as that seems to perform better than 64 bit. Binning prior to 25ns which gives a speed up over the previous setting.

Closes #86. 

Example of a simulated event's llh space with cross marking the true direction.
Before             |  After
:-------------------------:|:-------------------------:
Screen Shot 2022-11-30 at 14 08 38| Screen Shot 2022-11-30 at 14 08 27


tianluyuan:
A few more comparisons before/after the changes here. Removing all unhit DOMs caused local minima issues in the coarsest scan. Can add a sparse grid of unhit DOMs to mitigate this without much speed penalty. Also, seeding with both the fallback vertex position and the prior-nside vertex position improved the finer grid scans. Merging this into main, before the next icetray release is out can set `arguments = /cvmfs/icecube.opensciencegrid.org/users/tyuan/ubuntu_22.04/icetray.build/env-shell.sh ... ` in condor to use this reco.

Screen Shot 2022-12-11 at 10 59 38
Screen Shot 2022-12-11 at 11 00 18
Screen Shot 2022-12-11 at 11 01 16
Screen Shot 2022-12-11 at 11 02 30
Screen Shot 2022-12-11 at 11 03 35

PR #100, opened 2022-11-30
separate icetray-dependent util functions 

tianluyuan:
from non-icetray dependent ones. Obviates need for icetray to make plots.


ric-evans (approved):
We may want to call it `icetrayless` or something similar, so we know what makes these utils so simple.

PR #103, opened 2022-12-05
Import MQClient First

ric-evans:
```


dsschult (approved):

ric-evans:
The problem is better solved by https://github.com/icecube/skymap_scanner/pull/104

PR #104, opened 2022-12-06
new smaller Docker images with pulsar

dsschult:
Changes:
- manually compile pulsar in the docker image
- two-stage image build, using icetray-prod for the final image
- removed the download of the baseline_gcds - this is in the prod image
- removed the download of the extra spline tables - bind in yourself as needed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,9 @@
-FROM icecube/icetray:icetray-devel-v1.4.1-ubuntu22.04-2022-11-29 as build
+#
+# Define the base image icetray version
+#
+ARG ICETRAY_VERSION=v1.4.1-ubuntu22.04-2022-11-29
+
+FROM icecube/icetray:icetray-devel-$ICETRAY_VERSION as build
 
 #
 # Add Packages
@@ -23,7 +28,7 @@ RUN cd pulsar-client-python && git submodule update --init \
 #
 # Now make the prod image
 #
-FROM icecube/icetray:icetray-prod-v1.4.1-ubuntu22.04-2022-11-29 as prod 
+FROM icecube/icetray:icetray-prod-$ICETRAY_VERSION as prod
 
 
 #
@@ -86,8 +91,7 @@ RUN pip install .
 
 #
 # ENTRYPOINT
+# Inherited from icetray image as /usr/local/icetray/env-shell.sh
 #
 
-# set the entry point so that module is called with any parameters given to the `docker run` command
-#ENTRYPOINT ["/bin/bash", "/usr/local/icetray/env-shell.sh"]
 CMD []
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,17 +62,17 @@ RUN tree -f $I3_DATA
 RUN apt-get update 
 RUN DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends apt-utils
 RUN apt-get update 
-RUN DEBIAN_FRONTEND=noninteractive apt-get install -y bzip2 zstd default-jre protobuf-compiler python3-pybind11 python3-pip 
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y bzip2 zstd protobuf-compiler python3-pybind11 python3-pip 
 RUN apt-get clean
 
 #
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,76 +1,86 @@
...
-    ln -sf /usr/local/icetray/realtime_gfu/python /usr/local/icetray/lib/icecube/realtime_gfu
-# add realtime_hese
-RUN svn co http://code.icecube.wisc.edu/svn/meta-projects/realtime/releases/V21-06-00/realtime_hese \
-        /usr/local/icetray/realtime_hese --username=icecube --password=skua --no-auth-cache && \
-    ln -sf /usr/local/icetray/realtime_hese/python /usr/local/icetray/lib/icecube/realtime_hese
-# add realtime_tools
-RUN svn co http://code.icecube.wisc.edu/svn/meta-projects/realtime/releases/V21-06-00/realtime_tools \
-        /usr/local/icetray/realtime_tools --username=icecube --password=skua --no-auth-cache && \
-    ln -sf /usr/local/icetray/realtime_tools/python /usr/local/icetray/lib/icecube/realtime_tools
+FROM icecube/icetray:icetray-prod-v1.4.1-ubuntu22.04-2022-11-29 as prod
```
ric-evans, line 26: Does this tag need to be the same as the first `FROM`, as in `*v1.4.1-ubuntu22.04-2022-11-29`? If so, this may benefit by being a variable, so we don't have a version/tag mismatch in the future.

Dockerfile
```diff
@@ -79,5 +89,5 @@ RUN pip install .
 #
 
 # set the entry point so that module is called with any parameters given to the `docker run` command
-ENTRYPOINT ["/bin/bash", "/usr/local/icetray/env-shell.sh"]
+#ENTRYPOINT ["/bin/bash", "/usr/local/icetray/env-shell.sh"]
```
ric-evans, line 92: Why is this no longer needed?

skymap_scanner/recos/millipede_original.py
```diff
@@ -43,7 +43,7 @@ class MillipedeOriginal(RecoInterface):
     # At HESE energies, deposited light is dominated by the stochastic losses
     # (muon part emits so little light in comparison)
     # This is why we can use cascade tables
-    _splinedir = os.path.expandvars("$I3_TESTDATA/photospline")
+    _splinedir = os.path.expandvars("$I3_DATA/photon-tables/splines")
```
ric-evans, line 46: I'm a bit lost. How does this exist in CI if we are no longer bringing in the spline tables with the image build?

Dockerfile
dsschult: Not necessarily.  It's probably a good idea to keep them the same though.  I'm unsure how you would make this a variable in a Dockerfile.

Dockerfile
dsschult: As far as I can tell, if the ENTRYPOINT is not defined it defaults to the base image's ENTRYPOINT.  Which is exactly this.  Or at least the tests pass with this commented out?

skymap_scanner/recos/millipede_original.py
dsschult: The `prod` icetray image (the base image for this) installs a single spline table here, which just so happens to be the one we need for the original millipede.  Thus, CI works.

Dockerfile
ric-evans: I think it's similar to shell syntax `$myvar`: https://docs.docker.com/engine/reference/builder/#arg

tianluyuan:
Would it make sense to set an environment variable say `I3_DATA_CVMFS` pointing to the directory containing the other splines? That might be nicer than overwriting the default `I3_DATA` in env-shell or condor.

dsschult:
> Would it make sense to set an environment variable say `I3_DATA_CVMFS` pointing to the directory containing the other splines? That might be nicer than overwriting the default `I3_DATA` in env-shell or condor.

Maybe, but I'm not sure how it would be useful.  I think the better answer is passing the spline path through one of the client args, if we want to override it.  Then we can pick any directory we want.  This sounds like a job for another PR though.

Issue #79, opened 2022-10-31
Configurable N-Side Refinement Iterations


ric-evans:
@tianluyuan pointed out that the n-side iterations (refinements) are somewhat hardcoded currently (there's only a configurable min and max). A user-specified list would allow us to optimize this as needed, based on measurements we see in a production environment.

ric-evans:
https://github.com/icecube/skymap_scanner/pull/109 replaces `--max-nside` and `--min-nside` with a single `--nsides [min [max]]`. This also allows, defining a single nside value.

This is the first step toward closing this issue.

mlincett:
> #109 replaces `--max-nside` and `--min-nside` with a single `--nsides [min [max]]`. This also allows, defining a single nside value.
> 
> This is the first step toward closing this issue.

Thank you Ric! We should however define a way to limit the sky region for the scan with `nside=min`. Right now, my understanding is that `min-nside` is always an all-sky scan.

ric-evans:
> > #109 replaces `--max-nside` and `--min-nside` with a single `--nsides [min [max]]`. This also allows, defining a single nside value.
> > This is the first step toward closing this issue.
> 
> Thank you Ric! We should however define a way to limit the sky region for the scan with `nside=min`. Right now, my understanding is that `min-nside` is always an all-sky scan.

Sure. I'll have to rely on you or @tianluyuan to do that part :smile: SkyDriver will use a list of nside values, so from a systems perspective we'll be ready for the next steps

ric-evans:
Update: `--nsides` will now include pairings of the nside value and its pixel extension: `--nsides [min:pixext [max:pixext]]`

The goal:
```

Still a first step. I haven't touched anything beyond the argparse logic

ric-evans:
Note to self:

PR #109, opened 2022-12-19
Updates for SkyDriver

### LINKED ISSUE #95: Send Progress Reports & Results to SkyDriver REST API
After https://github.com/WIPACrepo/SkyDriver/issues/1, use SkyDriver's REST API to:
- send progress reports (which can eventually be forwarded to a 3rd party slack-posting script)
- send final results
   * we'll need to settle on a format for this, npz or json?

Removes/replaces all the "slack" logic in the scanner server. 

This method is more flexible--data will be persisted by SkyDriver's DB, and third parties can retrieve on a pull (GET). This opens possibilities for non-slack purposes 

Replaces https://github.com/icecube/skymap_scanner/issues/37

### LINKED ISSUE #108: Move Auth Tokens to Env Variables
Plain text tokens are better hidden by environment variables than CL arguments.

ric-evans:
Updates for SkyDriver's use. Changes are backward-compatible where/when needed.

#### Related Issues and PRs
includes https://github.com/Observation-Management-Service/ewms-pilot/pull/9
closes #95 
closes #108

#### Other Changes
- replaces `--max-nside` and `--min-nside` with a single `--nsides min [max]`. This also allows defining a single nside value
   + NOTE: so SkyDriver won't have to be updated after #79, the scanner will accept `--nsides min:pixext [max:pixext]` now. The pixel extension will be ignored, for the time being
   + see https://github.com/icecube/skymap_scanner/issues/79#issuecomment-1372887553
- replaces `--mini-test-variations` with `SKYSCAN_MINI_TEST=yes`
	+ this only applies to `millipede_original` test runs
- replaces `--timeout-to-clients` & `--timeout-from-clients` with `SKYSCAN_MQ_TIMEOUT_TO_CLIENTS` & `SKYSCAN_MQ_TIMEOUT_FROM_CLIENTS`
	+ if your starter scripts used these args, you'll need to update. If you didn't, the defaults have not changed
- replaces `--broker` with env var `SKYSCAN_BROKER_ADDRESS`


README.md
```diff
@@ -13,14 +13,18 @@ You will need to get a pulsar broker address and authentication token to pass to
 #### 1. Launch the Server
 The server can be launched from anywhere with a stable network connection. You can run it from the cobalts for example. For now, set `--timeout-to-clients` and `--timeout-from-clients` to a large value like 300000. This will persist the server in case the clients don't start up right away.
 ##### Figure Your Args
+###### Environment Variables
+```
+export SKYSCAN_BROKER_AUTH=$(cat ~/skyscan-broker.token)
```
dsschult, line 18: It's a bit weird that this is the only env variable, and it's really important.  Whereas the rest are cmd args, and several of those can be ignored.

ric-evans: We want to put auth tokens as env vars to obfuscate, right? I figure env vars are things that wouldn't need to change from one scan to another. 

I suppose we could move the skydriver and broker addresses, plus the timeouts, to env vars. What do you think?

dsschult: Yeah, that's a k8s requirement for obfuscation.  It might be good to make it clear why it's an env var (one-line comment or readme doc).

I'd be interested in moving the timeouts at least.

ric-evans: I think holding off on moving the timeout values, for now, is best since those values are still being tweaked--but I'll do the "env var as CL arg default" pattern.

ric-evans: Working on some README updates

ric-evans: > I think holding off on moving the timeout values, for now, is best since those values are still being tweaked--but I'll do the "env var as CL arg default" pattern.

On second thought, I'm going to change these to env vars. I prefer having one source of truth.

README.md
```diff
@@ -1,28 +1,38 @@
...
-The server can be launched from anywhere with a stable network connection. You can run it from the cobalts for example. For now, set `--timeout-to-clients` and `--timeout-from-clients` to a large value like 300000. This will persist the server in case the clients don't start up right away.
+The server can be launched from anywhere with a stable network connection. You can run it from the cobalts for example.
+
+###### In-Production Usage Note: timeout values
+For now, set `SKYSCAN_MQ_TIMEOUT_TO_CLIENTS` and `SKYSCAN_MQ_TIMEOUT_FROM_CLIENTS` to a large value like 300000. This will persist the server in case the clients don't start up right away.
+
 ##### Figure Your Args
+###### Environment Variables
+```
+export SKYSCAN_BROKER_AUTH=$(cat ~/skyscan-broker.token)  # obfuscated for security
```
tianluyuan, line 24: Is the broker auth tied to a particular broker address? If so it might not make sense to separate the two, otherwise it's fine.

README.md
```diff
@@ -106,15 +121,24 @@ This will pull all the events in the i3 file into `run*.evt*.json` which can be
...
+export SKYSCAN_MQ_TIMEOUT_TO_CLIENTS=300000
+export SKYSCAN_MQ_TIMEOUT_FROM_CLIENTS=300000
+ls /scratch/$USER/run*.condor | head -nN | xargs -I{} condor_submit {}
+```
 ```
 executable = /bin/sh 
-arguments = /usr/local/icetray/env-shell.sh python -m skymap_scanner.client --broker BROKER_ADDRESS --auth-token AUTH_TOKEN --timeout-to-clients 3600 --timeout-from-clients 3600 --startup-json-dir .
+arguments = /usr/local/icetray/env-shell.sh python -m skymap_scanner.client --broker BROKER_ADDRESS --startup-json-dir .
 +SingularityImage = "/cvmfs/icecube.opensciencegrid.org/containers/realtime/skymap_scanner:x.y.z"
+getenv = SKYSCAN_*
```
tianluyuan, line 141: Maybe better to just add to the `environment` variable inside the submit file, instead of breaking them up into `export` and `getenv`.

scripts/condor/spawn_condor_clients.py
```diff
@@ -172,9 +174,11 @@ def wait_for_file(path: str, wait_time: int, subpath: str = "") -> Path:
...
-        client_args = " ".join(f"{k} {v}" for k, v in json.load(f).items())
-        logging.info(f"Client Args: {client_args}")
+    client_args = ""
+    for carg_value in args.client_args:
+        carg, value = carg_value.split(":", maxsplit=1)
+        client_args += f" --{carg} {value} "
+    logging.info(f"Client Args: {client_args}")
     if "--startup-json-dir" in client_args:
         raise RuntimeError(
             "The '--client-args-json' file cannot include \"--startup-json-dir\". "
```
tianluyuan, line 184: Should this be `--client-args`?

tianluyuan (commented):
Have you tested through a full sequence of iterations to see if the new nside logic is working as expected?

ric-evans:
> Have you tested through a full sequence of iterations to see if the new nside logic is working as expected?

@tianluyuan Right now, the only implementation change is that the args are given as a pair of values, instead of individual CL arguments. It's still a min and max. See https://github.com/icecube/skymap_scanner/blob/d05328c176a1066121879f7fc998af4cb778ca2d/skymap_scanner/server/start_scan.py#L1105-L1113

README.md
ric-evans: > Is the broker auth tied to a particular broker address

ric-evans: > Is the broker auth tied to a particular broker address
Yes, as in the broker auth token is specific to a broker instance.

I could make the broker address an env var as well. My main objective is to put sensitive strings in env vars. The address is not necessarily sensitive info. But I see your point.

README.md
ric-evans: I'll check it out. I'll have to brush up on my condor

scripts/condor/spawn_condor_clients.py
ric-evans: good catch

README.md
ric-evans: I see what you mean now, I'll update the README

README.md
ric-evans: Seeing that you're our only manual user :smile: I've made them both env vars

PR #110, opened 2022-12-22
Deprecated files cleanup

mlincett:
Old codes are accessible from tag `v1.5` plus a minimal working interface to the old skymap scanner is provided [here](https://github.com/icecube/skymap-scanner-legacy).

Deprecated files can be dropped from main repo.


ric-evans:
Aha! I didn't see this... anyways: #120

PR #114, opened 2023-01-17
pin pulsar to a specific version

dsschult:


ric-evans (approved):
We may want to grab the most recent release, but not `main`. But since we want to move to rabbitmq, it's probably fine

dsschult:
> We may want to grab the most recent release, but not `main`. But since we want to move to rabbitmq, it's probably fine

Yeah, I figure it's good enough for now.  Rabbitmq won't need this build process, which is another advantage :smile:

PR #117, opened 2023-01-20
add script for generating a rabbitmq oauth2 token

dsschult:


dsschult:
I think it is now.  We tested it yesterday and got it working.

PR #118, opened 2023-01-23
Adding example scripts to start workers in k8s

briedel:


ric-evans:
This was integrated into SkyDriver ~v0.3.0

PR #122, opened 2023-01-26
Add `--real-event` & `--simulated-event` (required, mutex)

### LINKED ISSUE #7: skymap_scanner should take simulation or data as input without issue

ric-evans:
In order to distinguish scans of real events from simulated events, the user must now use either the `--real-event` or the `--simulated-event` flag. This metadata is stored in the `ScanResult` outputs and in SkyDriver's DB.

see https://github.com/icecube/skymap_scanner/issues/7#issuecomment-1404713746

closes #7


skymap_scanner/server/start_scan.py
```diff
@@ -1059,6 +1059,18 @@ def _nside_and_pixelextension(val: str) -> int:  # -> Tuple[int, int]:
         nargs='*',
         type=_nside_and_pixelextension,
     )
+    # --real-event XOR --simulated-event
+    group = parser.add_mutually_exclusive_group(required=True)
+    group.add_argument(
+        "--real-event",
```
tianluyuan, line 1065: Could we perhaps add an assertion for `--real-event` to ensure the run id is within some range? @kjmeagher has stated `real data runs are currently between 100000 and 137579`, so perhaps we can check between `100000-999999`.

ric-evans: Would it be sufficient to say `>=100000`?

mlincett (approved):
I am happy with the proposed solution!

skymap_scanner/server/start_scan.py
dsschult: actually, anything less than 1M is probably real data?  vs simulation is in the billions

tianluyuan: Yea I think there needs to be an upper bound since simulation runs can be greater. A lower bound could also be good since we know data will not go below while homebrew simulation may have runids like 0-999 or so.

ric-evans: since this will error loudly, we're not risking data corruption

tianluyuan: Yep that would be fine.

tianluyuan (approved):

PR #123, opened 2023-01-30
Client Starter [minor]

ric-evans:
A resource script for spawning clients on a condor cluster

see https://github.com/WIPACrepo/SkyDriver/issues/5


ric-evans:
fsdfsdf
ssdsd

ric-evans:
sdfsdf

resources/client_starter.py
```diff
@@ -0,0 +1,299 @@
...
+    LOGGER.info(job_description)
+
+    # dryrun?
+    if args.dryrun:
+        LOGGER.error("Script Aborted: Condor job not submitted")
+        return
+
+    # make connections -- do these before submitting so we don't have any unwanted surprises
+    skydriver_rc, scan_id = connect_to_skydriver()
+    schedd = get_schedd(args.collector_address, args.schedd_name)
```
dsschult, line 279: the args are called `collector` and `schedd`.  maybe they need to be renamed?

dsschult (changes requested):
Generally looks good.  My only comment is that it would be nice to have a mode where the collector/schedd was filled in from local host values, so it could connect automatically for someone like Tianlu.  But it would probably work as-is.

ric-evans:
> Generally looks good. My only comment is that it would be nice to have a mode where the collector/schedd was filled in from local host values, so it could connect automatically for someone like Tianlu. But it would probably work as-is.

I'm thinking that's what will happen when the CL args are not given (`""`)

dsschult (approved):

PR #126, opened 2023-02-09
Replace `--startup-json-dir` with `--client-startup-json`

ric-evans:
This simplifies the relationship between the client starter and the scanner. 

For manual running: This change also allows for uniquely named startup-json files without the need for uniquely-named parent directories.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -117,7 +117,7 @@ jobs:
             --simulated-event \
             &
 
-          ./resources/launch_scripts/wait_for_file.sh ./startup.json CLIENT_STARTER_WAIT_FOR_STARTUP_JSON
+          ./resources/launch_scripts/wait_for_file.sh ./startup.json $CLIENT_STARTER_WAIT_FOR_STARTUP_JSON
 
           # Launch Clients
           nclients=$(( $CLIENTS_PER_CPU * $(nproc) ))
@@ -198,7 +198,7 @@ jobs:
             --real-event \
             &
 
-          ./resources/launch_scripts/wait_for_file.sh ./startup.json CLIENT_STARTER_WAIT_FOR_STARTUP_JSON
+          ./resources/launch_scripts/wait_for_file.sh ./startup.json $CLIENT_STARTER_WAIT_FOR_STARTUP_JSON
 
           # Launch Clients
           nclients=$(( $CLIENTS_PER_CPU * $(nproc) ))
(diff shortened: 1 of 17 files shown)


ric-evans:
@tianluyuan I attempted to update your notes in README. I'm not too familiar with `xargs`, so you may want to take a look

tianluyuan:
Is there a better name than `startup-json-file`?

I think part of what makes this argument and `--cache-dir` confusing is that it seems like the user shouldn't need to know about them. Everything that uniquely defines an event reconstruction is specified by the other arguments as in `python -m skymap_scanner.server --reco-algo millipede_wilks --event-file EVENT.json [--nsides ...]`. It would be nice if there was a way to hide this from users.

dsschult:
> Is there a better name than `startup-json-file`?
> 
> I think part of what makes this argument and `--cache-dir` confusing is that it seems like the user shouldn't need to know about them. Everything that uniquely defines an event reconstruction is specified by the other arguments as in `python -m skymap_scanner.server --reco-algo millipede_wilks --event-file EVENT.json [--nsides ...]`. It would be nice if there was a way to hide this from users.

I guess this is needed to transfer information to the clients.  Maybe `--client-startup-file`?

The real way to hide this from users is SkyDriver, which hides server and client startup and management.

ric-evans:
> > Is there a better name than `startup-json-file`?
> > I think part of what makes this argument and `--cache-dir` confusing is that it seems like the user shouldn't need to know about them. Everything that uniquely defines an event reconstruction is specified by the other arguments as in `python -m skymap_scanner.server --reco-algo millipede_wilks --event-file EVENT.json [--nsides ...]`. It would be nice if there was a way to hide this from users.
> 
> I guess this is needed to transfer information to the clients. Maybe `--client-startup-file`?
> 
> The real way to hide this from users is SkyDriver, which hides server and client startup and management.

Let's go with `--client-startup-json`--that nails what each part of the argument is.

And yeah, ultimately the user won't have to know about these arguments.

mlincett:
> And yeah, ultimately the user won't have to know about these arguments.

Agreed. But from development (for example, implementation of a new reco) one is still expected to deal with inputs manually, unless SkyDriver has a path for the selection of a custom singularity image when running the scan.

ric-evans:
> > And yeah, ultimately the user won't have to know about these arguments.
> 
> Agreed. But from development (for example, implementation of a new reco) one is still expected to deal with inputs manually, unless SkyDriver has a path for the selection of a custom singularity image when running the scan.

Actually, it will! You'll be able to pass in any skymap_scanner docker tag that's available on CVMFS

PR #128, opened 2023-02-14
Use `oms-mqclient`

ric-evans:
Most notably, this restores the functionality of `SKYSCAN_BROKER_AUTH`. We no longer need to prepend the token string to the address url.


.github/workflows/tests.yml
```diff
@@ -102,6 +99,8 @@ jobs:
           ls -lh skymap_scanner.sif
       - name: Run Singularity Container
         run: |
+          ./MQClient/resources/docker-rabbitmq.sh $BROKER_CONTAINER_NAME
```
dsschult, line 102: Note that I am personally against manually running docker containers, unless there's a good reason to.  When you run it as a service Actions automatically saves the logs, instead of a somewhat hacky way of manually getting the logs.

Running services instead of docker manually also signals that you have "good design choices" and don't need to override things to get it working.

ric-evans: Sure, I see your point. My main reasoning was to get this working with a setup I know works, then prune it back to a `services` container if manageable. I'm not sure how to do the `custom.conf` binding, I guess I'd need to persist a file on GH.

dsschult: Technically with services you can't do a file binding.  I'm unsure it's really necessary, and stems from our new inability to use other usernames than `guest` (and that not working on non-localhost).

PR #129, opened 2023-02-14
K8s setup in gcp

briedel:


dsschult:
I think #128 should come first.  once that's merged in, these tests should pass again

Issue #130, opened 2023-02-15
In Test: Use `services` for RabbitMQ Server


ric-evans:
Technically with services you can't do a file binding.  I'm unsure it's really necessary, and stems from our new inability to use other usernames than `guest` (and that not working on non-localhost).

_Originally posted by @dsschult in https://github.com/icecube/skymap_scanner/pull/128#discussion_r1107297493_

PR #131, opened 2023-02-15
In Test: Use `services` for RabbitMQ Server

ric-evans:
~This is a long one:
In order to use an auth-less rabbitmq, we must use `guest/guest`. This is the default that mqclient uses when no `auth_token` value is provided (which also prevents us from using other `username/password` combos). In order to use `guest/guest` we must be running the rabbitmq server on a local network. Normally, that could be done using `--network=host`, but that's not allowed in GHA's `services` (it's explicitly forbidden for some reason). So, we have to configure the rabbitmq server using a `custom.conf`. However, we cannot mount/bind files in the repo to the container using `services`. The solution is to bind a staging directory in the runner. Then in a `steps.run` write/mount the file, manually transfer the `custom.conf` contents, and do a `docker restart`.~

Using the update in https://github.com/Observation-Management-Service/MQClient/pull/85, we can now use auth with a username and password with GHA's `services`. In production, we'll use a token provided by the Keycloak service.

closes https://github.com/icecube/skymap_scanner/issues/130


ric-evans:
asdasdasd

dsschult (approved):

PR #133, opened 2023-02-16
Client starter fixes

mlincett:
- support for the new structure of the "startup JSON" argument
- make additional client arguments optional

Maybe some more polishing can be done before merging but this works (at least with "dummy" reco).


mlincett:
Nevermind, it seems I did not branch from the latest main (but I could swear I fetched/pulled whatever).

PR #134, opened 2023-02-17
Fixes to client starter

mlincett:
- make client args optional
- add requirements for `millipede_wilks` (AVX, CVMFS access, I3DATA env variables in Singularity)

### DIFF
--- a/resources/client_starter.py
+++ b/resources/client_starter.py
@@ -71,15 +71,14 @@ def make_condor_job_description(  # pylint: disable=too-many-arguments
 
     # Build the environment specification for condor
     vars = []
-    #
+    # RABBITMQ_* and EWMS_* are inherited via condor `getenv`, but we have default in case these are not set.
     if not os.getenv("RABBITMQ_HEARTBEAT"):
         vars.append("RABBITMQ_HEARTBEAT=600")
     if not os.getenv("EWMS_PILOT_QUARANTINE_TIME"):
         vars.append("EWMS_PILOT_QUARANTINE_TIME=1800")
-
+    # The container sets I3_DATA to /opt/i3-data, however `millipede_wilks` requires files (spline tables) that are not available in the image. For the time being we require CVFMS and we load I3_DATA from there. In order to override the environment variables we need to prepend APPTAINERENV_ or SINGULARITYENV_ to the variable name. There are site-dependent behaviour but these two should cover all cases. See https://github.com/icecube/skymap_scanner/issues/135#issuecomment-1449063054.
     for prefix in ["APPTAINERENV_", "SINGULARITYENV_"]:
         vars.append(f"{prefix}I3_DATA=/cvmfs/icecube.opensciencegrid.org/data")
-
     environment = " ".join(vars)
 
     # write
--- a/resources/client_starter.py
+++ b/resources/client_starter.py
@@ -229,7 +229,7 @@ def main() -> None:
     )
     parser.add_argument(
         "--client-args",
-        required=True,
+        required=False,
         nargs="+",
         help="n 'key:value' pairs containing the python CL arguments to pass to skymap_scanner.client",
     )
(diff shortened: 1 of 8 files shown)


resources/client_starter.py
```diff
@@ -69,20 +69,27 @@ def make_condor_job_description(  # pylint: disable=too-many-arguments
...
     # write
     submit_dict = {
         "executable": "/bin/sh",
         "arguments": f"/usr/local/icetray/env-shell.sh python -m skymap_scanner.client {client_args} --client-startup-json ./{client_startup_json.name}",
         "+SingularityImage": singularity_image,
+        "Requirements": "HAS_CVMFS_icecube_opensciencegrid_org && has_avx",
         "getenv": "SKYSCAN_*, EWMS_*, RABBITMQ_*, PULSAR_UNACKED_MESSAGES_TIMEOUT_SEC",
         "output": str(logs_subdir / "client-$(ProcId).out"),
-        "environment": environment,
+        "environment": f'"{environment}"',
```
ric-evans, line 92: Are the quotes needed here?

resources/client_starter.py
```diff
@@ -253,15 +260,16 @@ def wait_for_file(waitee: Path, wait_time: int) -> Path:
...
+    if args.client_args is not None:
+        for carg_value in args.client_args:
+            carg, value = carg_value.split(":", maxsplit=1)
+            client_args += f" --{carg} {value} "
+            LOGGER.info(f"Client Args: {client_args}")
+            if "--client-startup-json" in client_args:
+                raise RuntimeError(
+                    "The '--client-args' arg cannot include \"--client-startup-json\". "
+                    "This needs to be given to this script explicitly ('--client-startup-json')."
+                )
```
ric-evans, line 272: Looks like this was indented one too many levels

resources/client_starter.py
```diff
@@ -69,20 +69,27 @@ def make_condor_job_description(  # pylint: disable=too-many-arguments
...
+
     if not os.getenv("RABBITMQ_HEARTBEAT"):
-        environment = f"RABBITMQ_HEARTBEAT=600 {environment}"
+        vars.append("RABBITMQ_HEARTBEAT=600")
     if not os.getenv("EWMS_PILOT_QUARANTINE_TIME"):
-        environment = f"EWMS_PILOT_QUARANTINE_TIME=1800 {environment}"
+        vars.append("EWMS_PILOT_QUARANTINE_TIME=1800")
+
+    for prefix in ["APPTAINERENV_", "SINGULARITYENV_"]:
+        vars.append(f"{prefix}I3_DATA=/cvmfs/icecube.opensciencegrid.org/data")
```
ric-evans, line 80: Can you add a comment explaining why this is happening, or even a URL to https://github.com/icecube/skymap_scanner/issues/135#issuecomment-1449063054?

resources/client_starter.py
```diff
@@ -69,20 +69,27 @@ def make_condor_job_description(  # pylint: disable=too-many-arguments
...
+
+    environment = " ".join(vars)
```
ric-evans, line 82: You could use this `join()` directly in the actual dictionary assignment below. But it's not a deal breaker

ric-evans (changes requested):

resources/client_starter.py
mlincett: Maybe? this comes out of `black`.

ric-evans: Oh, I'm not talking about style. When it logs `client_args` & checks for `"--client-startup-json"`, this should be outside the for-loop--look at the diff.

ric-evans (changes requested):

resources/client_starter.py
mlincett: Ah, right. Sorry, I had tunnel-vision.

resources/client_starter.py
mlincett: True, but conceptually I like the separation. If in the future we decide to build the "environment" string differently we do not have to change the code in the submit_dict block.

resources/client_starter.py
ric-evans: Sure, condor does what condor wants, so who knows!

ric-evans (approved):
Thanks for your hard work here!

PR #146, opened 2023-03-03
Fix CI for Bots on `main` Pt-5

ric-evans:
Now we're using the GitHub Checks API


ric-evans:
A few quirks of the API:
- a job becomes a "check" once it has been created for a commit, meaning if Y job depends on X, and X hasn't finished, then the check for Y does not exist. So any query about check Y will come back empty
- canceled jobs are considered "completed" checks
- if a job is re-run, then queries will see only the previous run's check until the job's check is re-created

PR #147, opened 2023-03-03
Simplification and improvement of GCD handling

mlincett:
Up to now:
- removal of the `config.GCD_BASE_DIRS` list, assuming GCD files are always read from the container `DEFAULT_GCD_DIR`;
- simplification of GCD handling in `extract_json_message()` and `load_cache_state()`
- improvement to the naming of variables and functions, in particular `GCD_diff_base_*` is now consistently named `baseline_GCD_*`
- avoid reusing the same variable name for different purposes along the code;
- some hardcoded strings have been parameterized in `config.py`;
- comments and documentation;
- removal of IceTray filestagers for reading GCD files.

### DIFF
--- a/skymap_scanner/client/reco_icetray.py
+++ b/skymap_scanner/client/reco_icetray.py
@@ -87,35 +87,29 @@ def get_baseline_GCD_handle(baseline_GCD_file: str) -> Any:
     """Find an available GCD base path."""
     stagers = dataio.get_stagers()
 
-    LOGGER.debug(f"Look up baseline GCD at {baseline_GCD_file}")
+    LOGGER.debug(f"Baseline GCD at {baseline_GCD_file}")
 
-    # try to load the base file from the various possible input directories
-    GCD_diff_base_handle = None
+    # assuming baseline_GCD_file is a valid GCD path!
+    # NOTE: old code use to search for a basename in a list of GCD_BASE_DIRS
+    baseline_GCD_handle = None
     if baseline_GCD_file not in [None, "None"]:
-        for GCD_base_dir in cfg.GCD_BASE_DIRS:
-            try:
-                read_url = os.path.join(GCD_base_dir, baseline_GCD_file)
-                LOGGER.debug("reading baseline GCD from {0}".format(read_url))
-                GCD_diff_base_handle = stagers.GetReadablePath(read_url)
-                if not os.path.isfile(str(GCD_diff_base_handle)):
-                    raise RuntimeError("file does not exist (or is not a file)")
-            except:
-                LOGGER.debug(" -> failed")
-                GCD_diff_base_handle = None
-            if GCD_diff_base_handle is not None:
-                LOGGER.debug(" -> success")
-                break
+        try:
+            LOGGER.debug("reading baseline GCD from {0}".format(baseline_GCD_file))
+            baseline_GCD_handle = stagers.GetReadablePath(baseline_GCD_file)
+            if not os.path.isfile(str(baseline_GCD_handle)):
+                raise RuntimeError("file does not exist (or is not a file)")
+        except:
+            LOGGER.debug(" -> failed")
+            baseline_GCD_handle = None
+        if baseline_GCD_handle is not None:
+            LOGGER.debug(" -> success")
 
-        if GCD_diff_base_handle is None:
-            raise RuntimeError(
-                "Could not read the input GCD file '{0}' from any pre-configured location".format(
-                    baseline_GCD_file
-                )
-            )
+    if baseline_GCD_handle is None:
+        raise RuntimeError(f"Could not read the input GCD file '{baseline_GCD_file}'")
     else:
-        LOGGER.info("baseline_GCD_file is None! Is this expected?")
+        LOGGER.info("baseline_GCD_file is None. Likely frame packet is not a GCD diff.")
 
-    return GCD_diff_base_handle
+    return baseline_GCD_handle
 
 
 def reco_pixel(
--- a/skymap_scanner/client/reco_icetray.py
+++ b/skymap_scanner/client/reco_icetray.py
@@ -125,22 +125,13 @@ def reco_pixel(
     tray.AddModule(notifyStart, "notifyStart")
 
     # get GCD
-    @icetray.traysegment
-    def UncompressGCD(tray, name, base_GCD_path, base_GCD_filename):
-        tray.Add(
-            uncompress,
-            name + "_GCD_patch",
-            keep_compressed=False,
-            base_path=base_GCD_path,
-            base_filename=base_GCD_filename,
-        )
-
     if check_baseline_GCD(baseline_GCD_file):
         tray.Add(
-            UncompressGCD,
-            "GCD_uncompress",
-            base_GCD_path="", 
-            base_GCD_filename=baseline_GCD_file,
+            uncompress,
+            "GCD_uncompress_GCD_patch",
+            keep_compressed=False,
+            base_path="",
+            base_filename=baseline_GCD_file
         )
 
     # perform fit
--- a/skymap_scanner/client/reco_icetray.py
+++ b/skymap_scanner/client/reco_icetray.py
@@ -8,7 +8,7 @@ import logging
 import os
 import pickle
 from pathlib import Path
-from typing import Any, List
+from typing import Any, List, Union
 
 from I3Tray import I3Tray  # type: ignore[import]
 from icecube import (  # type: ignore[import]  # noqa: F401
@@ -83,34 +83,16 @@ def save_to_disk_cache(frame: icetray.I3Frame, save_dir: Path) -> Path:
     return pixel_fname
 
 
-def get_baseline_GCD_handle(baseline_GCD_file: str) -> Any:
-    """Find an available GCD base path."""
-    stagers = dataio.get_stagers()
+def check_baseline_GCD(baseline_GCD_file : Union[str, None]) -> bool:
+    LOGGER.debug(f"Testing baseline GCD at {baseline_GCD_file}")
 
-    LOGGER.debug(f"Baseline GCD at {baseline_GCD_file}")
-
-    # assuming baseline_GCD_file is a valid GCD path!
-    # NOTE: old code use to search for a basename in a list of GCD_BASE_DIRS
-    baseline_GCD_handle = None
-    if baseline_GCD_file not in [None, "None"]:
-        try:
-            LOGGER.debug("reading baseline GCD from {0}".format(baseline_GCD_file))
-            baseline_GCD_handle = stagers.GetReadablePath(baseline_GCD_file)
-            if not os.path.isfile(str(baseline_GCD_handle)):
-                raise RuntimeError("file does not exist (or is not a file)")
-        except:
-            LOGGER.debug(" -> failed")
-            baseline_GCD_handle = None
-        if baseline_GCD_handle is not None:
-            LOGGER.debug(" -> success")
-
-    if baseline_GCD_handle is None:
-        raise RuntimeError(f"Could not read the input GCD file '{baseline_GCD_file}'")
+    if baseline_GCD_file is None:
+        return False
+    
+    if os.path.isfile(baseline_GCD_file):
+        return True
     else:
-        LOGGER.info("baseline_GCD_file is None. Likely frame packet is not a GCD diff.")
-
-    return baseline_GCD_handle
-
+        raise RuntimeError(f"The provided baseline GCD file could not be read '{baseline_GCD_file}'. Cannot build GCD from GCD diff.")
 
 def reco_pixel(
     reco_algo: str,
@@ -153,12 +135,12 @@ def reco_pixel(
             base_filename=base_GCD_filename,
         )
 
-    if baseline_GCD_handle := get_baseline_GCD_handle(baseline_GCD_file):
+    if check_baseline_GCD(baseline_GCD_file):
         tray.Add(
             UncompressGCD,
             "GCD_uncompress",
-            base_GCD_path="",
-            base_GCD_filename=str(baseline_GCD_handle),
+            base_GCD_path="", 
+            base_GCD_filename=baseline_GCD_file,
         )
 
     # perform fit
--- a/skymap_scanner/client/reco_icetray.py
+++ b/skymap_scanner/client/reco_icetray.py
@@ -185,11 +185,9 @@ def reco_pixel(
                 f"Pickle-dumping reco {pixelreco.pixel_to_tuple(frame)}: "
                 f"{frame_for_logging(frame)} to {out_pkl}."
             )
-            # apparently baseline GCD is sufficient here, maybe filestager can be None
             geometry = get_baseline_gcd_frames(
                 baseline_GCD_file,
-                GCDQp_packet,
-                filestager=dataio.get_stagers(),
+                GCDQp_packet
             )[0]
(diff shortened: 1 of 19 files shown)


skymap_scanner/server/start_scan.py
```diff
@@ -114,7 +114,7 @@ def __init__(
         self.reporter_start_time = 0.0
 
         self._call_order = {
-            'current_previous': {  # current_fucntion: previous_fucntion(s)
+            "current_previous": {  # current_fucntion: previous_fucntion(s)
```
ric-evans, line 117: I see you've started using Black :smile: One thing I've done to keep quote-conversions down is use `--skip-string-normalization` "smartly": `black $(if [[ $(grep -o -P '#\\s*fmt:\\s*quotes-ok*' $FILE) && $(grep -o -P '\\x27[^\\x27]+\\x27' $FILE | wc -l) -gt 10 ]]; then echo '--skip-string-normalization'; fi) $FILE`. This adds the benefit of keeping `git blame` true to the author but also doesn't fully exclude quote-conversions in newer/smaller files.

ric-evans (commented):
I generally like the direction this is going. Hopefully, this eliminates tech debt. It looks like you have a few more things to do, so let me know when you've finished that and I can give it a more thorough review.

skymap_scanner/server/start_scan.py
mlincett: Yes, sorry, I have it run automatically on save. This file can simply be reverted.

mlincett:
> I generally like the direction this is going. Hopefully, this eliminates tech debt. It looks like you have a few more things to do, so let me know when you've finished that and I can give it a more thorough review.

Actually I thought I was finished. But indeed there are still a couple of places where I can remove filestagers.

ric-evans:
> Work in progress / to do:
> 
> * removal of IceTray filestagers for reading GCD files.

Oh okay. I was referring to:
> Work in progress / to do:
> removal of IceTray filestagers for reading GCD files.
All to your discretion what gets done in this PR versus a future one.

skymap_scanner/server/start_scan.py
ric-evans: Are you using Sublime? If so, I can give you some on-save code that does this automatically

skymap_scanner/client/reco_icetray.py
```diff
@@ -83,35 +83,33 @@ def save_to_disk_cache(frame: icetray.I3Frame, save_dir: Path) -> Path:
...
-            raise RuntimeError(
-                "Could not read the input GCD file '{0}' from any pre-configured location".format(
-                    baseline_GCD_file
-                )
-            )
+        try:
+            LOGGER.debug("reading baseline GCD from {0}".format(baseline_GCD_file))
+            baseline_GCD_handle = stagers.GetReadablePath(baseline_GCD_file)
+            if not os.path.isfile(str(baseline_GCD_handle)):
+                raise RuntimeError("file does not exist (or is not a file)")
```
ric-evans, line 100: may be nice to include the filename

skymap_scanner/client/reco_icetray.py
```diff
@@ -83,35 +83,33 @@ def save_to_disk_cache(frame: icetray.I3Frame, save_dir: Path) -> Path:
...
+        except:
+            LOGGER.debug(" -> failed")
+            baseline_GCD_handle = None
+        if baseline_GCD_handle is not None:
+            LOGGER.debug(" -> success")
+
+    if baseline_GCD_handle is None:
+        raise RuntimeError(f"Could not read the input GCD file '{baseline_GCD_file}'")
+    else:
+        LOGGER.info("baseline_GCD_file is None. Likely frame packet is not a GCD diff.")
```
ric-evans, line 110: I guess I don't follow the logic that `baseline_GCD_handle is None` implies `baseline_GCD_file is None`, at least without intense reading.

skymap_scanner/client/reco_icetray.py
```diff
@@ -187,11 +185,9 @@ def writeout_reco(frame: icetray.I3Frame) -> None:
                 f"Pickle-dumping reco {pixelreco.pixel_to_tuple(frame)}: "
                 f"{frame_for_logging(frame)} to {out_pkl}."
             )
-            # apparently baseline GCD is sufficient here, maybe filestager can be None
```
ric-evans, line 190: :smile: nice

skymap_scanner/client/reco_icetray.py
```diff
@@ -155,12 +153,12 @@ def UncompressGCD(tray, name, base_GCD_path, base_GCD_filename):
...
         )
 
-    if GCD_diff_base_handle := get_GCD_diff_base_handle(baseline_GCD_file):
+    if baseline_GCD_handle := get_baseline_GCD_handle(baseline_GCD_file):
         tray.Add(
             UncompressGCD,
             "GCD_uncompress",
             base_GCD_path="",
-            base_GCD_filename=str(GCD_diff_base_handle),
+            base_GCD_filename=str(baseline_GCD_handle),
```
ric-evans, line 161: Question: is "base" just short for "baseline"?

skymap_scanner/utils/extract_json_message.py
```diff
@@ -37,11 +37,13 @@ def extract_GCD_diff_base_filename(frame_packet):
...
                 GCD_diff_base_filename = frame[key].base_filename
+                LOGGER.debug(f"GCD diff base_filename loaded from {key} in {frame.Stop} frame.")
             elif frame[key].base_filename != GCD_diff_base_filename:
                 raise RuntimeError("inconsistent frame diff base GCD file names. expected {0}, got {1}".format(GCD_diff_base_filename, frame[key].base_filename))
 
     if GCD_diff_base_filename == "current-gcd":
-        LOGGER.warning(" **** WARNING: baseline GCD file is \"current-gcd\". replacing with \"2016_01_08_Run127381.i3\".")
+        # It is unclear which legacy case is covered by this condition.
+        LOGGER.warning("Baseline GCD file is \"current-gcd\". Replacing with \"2016_01_08_Run127381.i3\".")
         GCD_diff_base_filename = "2016_01_08_Run127381.i3"
```
ric-evans, line 47: this is concerning, but perhaps we can't avoid it

skymap_scanner/utils/extract_json_message.py
```diff
@@ -135,85 +134,77 @@ def __extract_frame_packet(
...
+        with open(source_baseline_GCD_metadata, "w") as text_file:
+            text_file.write(baseline_GCD)   
+
+    #=====================
+    # GCD-less framepacket
+    #=====================
+    if ("I3Geometry" not in frame_packet[0]) and ("I3GeometryDiff" not in frame_packet[0]):
+        LOGGER.debug("Packet with empty GCD frames. Need to load baseline GCD")
+        # If no GCD is specified, work out correct one from run number
+        available_GCDs = sorted([x for x in os.listdir(GCD_dir) if ".i3" in x])
```
ric-evans, line 207: That's a lot of diff--I'll have to trust you here. Perhaps @tianluyuan could give a better review of this change.

skymap_scanner/utils/load_scan_state.py
```diff
@@ -124,86 +109,71 @@ def load_scan_state(
...
+                read_path = os.path.join(cfg.DEFAULT_GCD_DIR, source_baseline_GCD_basename)
+                LOGGER.debug("load_GCDQp_state => reading source baseline GCD from {0}".format(read_path))
+                source_baseline_GCD_framepacket = load_framepacket_from_file(read_path)
+            except:
+                LOGGER.debug(" -> failed")
+                source_baseline_GCD_framepacket = None
+            if source_baseline_GCD_framepacket is None:
+                raise RuntimeError(f"load_GCDQp_state => Could not read the source GCD file \"{source_baseline_GCD_metadata}\"")
+            
+            cached_baseline_GCD_framepacket = load_framepacket_from_file(cached_baseline_GCD)
```
ric-evans, line 160: It's too bad Githiub doesn't see the commonalities in these lines--I think this section looks fine. I suppose that's why we have tests too

skymap_scanner/utils/prepare_frames.py
```diff
@@ -110,6 +110,7 @@ def copyPulseName(frame, old_name, new_name):
         If=lambda frame: nominalPulsesName+'HLC' not in frame)
 
     if reco_algo.lower() == 'millipede_original':
+        # TODO: documentation for this conditional statement
```
ric-evans, line 113: Found a sneaky `TODO`

skymap_scanner/client/reco_icetray.py
```diff
@@ -83,35 +83,33 @@ def save_to_disk_cache(frame: icetray.I3Frame, save_dir: Path) -> Path:
     return pixel_fname
 
 
-def get_GCD_diff_base_handle(baseline_GCD_file: str) -> Any:
+def get_baseline_GCD_handle(baseline_GCD_file: str) -> Any:
     """Find an available GCD base path."""
     stagers = dataio.get_stagers()
```
tianluyuan, line 88: Is the stager here needed still?

skymap_scanner/utils/extract_json_message.py
```diff
@@ -224,48 +215,49 @@ def __extract_frame_packet(
...
-        frame_packet[0] = ehe_override_gcd[0]
-        frame_packet[1] = ehe_override_gcd[1]
-        frame_packet[2] = ehe_override_gcd[2]
-        del ehe_override_gcd
-
-    if GCD_diff_base_filename is not None:
-        frame_packet = prepare_frames(frame_packet, str(GCD_diff_base_handle), reco_algo, pulsesName=pulsesName)
+        LOGGER.debug(f"By process of elimination using run numbers, using {baseline_GCD_file}")
+        # NOTE: commented because it should never occur
+        # if baseline_GCD_file is None:
```
tianluyuan, line 223: perhaps should just delete this?

skymap_scanner/utils/extract_json_message.py
tianluyuan: How does this actually pick out the GCD file for a particular GCD-less event? It looks like it will use the latest baseline GCD file that it can find, but that may not (probably does not) correspond to the actual GCD of the event?

I guess I am saying it seems like all frame packets should have at least a GCD-diff.

tianluyuan (commented):
Thanks for going through the GCD simplification. I think this looks good and definitely better than before. I left a few comments inline, nothing too major.

skymap_scanner/config.py
```diff
@@ -47,6 +36,10 @@
 MSG_KEY_RECO_ALGO: Final = "reco_algo"
 MSG_KEY_PFRAME: Final = "pframe"
 
+BASELINE_GCD_FILENAME = "base_GCD_for_diff.i3"
+SOURCE_BASELINE_GCD_METADATA = "original_base_GCD_for_diff_filename.txt"
+GCDQp_FILENAME = "GCDQp.i3"
```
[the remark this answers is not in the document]
ric-evans: Personally, I do prefer a catch-all `config.py` for things that are core to the application's functionality--especially env vars. But I could see a few good reasons for additional, specific `constants.py` files, like `server/constants.py`.

skymap_scanner/client/reco_icetray.py
mlincett: "diff_base" stands for "base GCD from which the GCD diff is calculated". This is equivalent with what we call "baseline GCD" otherwise.

skymap_scanner/utils/extract_json_message.py
mlincett: It seems that this applies only to very old alerts. It is very realtime-specific so I expect this will be decoupled from Skymap Scanner in the future.

skymap_scanner/utils/extract_json_message.py
mlincett: Correct @tianluyuan . According to documentation this logic applies to old-format EHE alerts, but I found out it will also trigger if you pass as an input subthreshold GFU event (that is GCD-less). For the time being I think is good to support this scenario in the code (although with all possible warnings).

skymap_scanner/utils/prepare_frames.py
mlincett: This is for @tianluyuan I guess. Can you provide a brief comment about how this preprocessing is different for "millipede original" and "millipede wilks"?

skymap_scanner/config.py
```diff
@@ -14,17 +14,6 @@
...
-GCD_BASE_DIRS: Final = [
-    os.path.join(
-        os.environ["HOME"], "PoleBaseGCDs"
-    ),  # why can't we reach anything from the followup nodes???
-    "file:///data/user/followup/baseline_gcds",
-    "http://icecube:skua@convey.icecube.wisc.edu/data/user/followup/baseline_gcds",
-    "file:///data/exp/IceCube/2016/internal-system/PoleBaseGCDs",
-    "http://icecube:skua@convey.icecube.wisc.edu/data/exp/IceCube/2016/internal-system/PoleBaseGCDs",
-    "file:///cvmfs/icecube.opensciencegrid.org/users/steinrob/GCD/PoleBaseGCDs/",
-]
```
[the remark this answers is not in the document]
mlincett: That's the current solution, the baseline GCDs are fetched when the image is built.

skymap_scanner/config.py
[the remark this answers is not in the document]
mlincett: I am flexible about this, I just wanted a place to put those for the time being.

All in all this is also legacy functionality to allow the scan to resume from a previously existing cache, but it is not very useful in the new design.

skymap_scanner/client/reco_icetray.py
```diff
@@ -83,35 +83,33 @@ def save_to_disk_cache(frame: icetray.I3Frame, save_dir: Path) -> Path:
...
-                break
-
-        if GCD_diff_base_handle is None:
-            raise RuntimeError(
-                "Could not read the input GCD file '{0}' from any pre-configured location".format(
-                    baseline_GCD_file
-                )
-            )
+        try:
+            LOGGER.debug("reading baseline GCD from {0}".format(baseline_GCD_file))
```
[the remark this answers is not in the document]
mlincett: The purpose of this code is to check whether a baseline GCD is provided, so the "GCD uncompress" IceTray segment can be added to the tray. Now I just check whether `baseline_GCD_file` points to a valid file.

skymap_scanner/config.py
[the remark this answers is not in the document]
mlincett: I suggest we move this to a new issue and decouple the discussion from this PR.

skymap_scanner/client/reco_icetray.py
```diff
@@ -83,35 +83,33 @@ def save_to_disk_cache(frame: icetray.I3Frame, save_dir: Path) -> Path:
...
     stagers = dataio.get_stagers()
 
-    # try to load the base file from the various possible input directories
-    GCD_diff_base_handle = None
+    LOGGER.debug(f"Baseline GCD at {baseline_GCD_file}")
+
+    # assuming baseline_GCD_file is a valid GCD path!
+    # NOTE: old code use to search for a basename in a list of GCD_BASE_DIRS
+    baseline_GCD_handle = None
     if baseline_GCD_file not in [None, "None"]:
```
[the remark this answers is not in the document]
mlincett: Hopefully not.

tianluyuan (approved):
Thanks looks good.

PR #152, opened 2023-03-09
Move Client Starter

ric-evans:
The client starter script has moved to github.com/WIPACrepo/SkyDriver/blob/main/clientmanager/client_starter.py

It has more in common with SkyDriver than the scanner proper.


dsschult:
My opinion: we might want to leave this here until we're confident in using SkyDriver.  This is at least an interim way to launch all the clients.

ric-evans:
We'll remove this later

Issue #155, opened 2023-03-21
Progress Report Updates


ric-evans:
1. make updates following #79 (https://github.com/icecube/skymap_scanner/issues/79#issuecomment-1374143776)
2. ~update `prior processing = time of first reco - global start time`. Since we can't know how long the actual start-up takes, we have to assume a majority of this "first reco time" is build time & resource gathering. This way, we can measure reco speed more accurately.~ _This is an interesting problem, it would best be solved by the clients sending an initial "I'm here" message. The mq-pub may be better suited as part of the EWMS pilot, than the client proper._ **Update:** This was solved by #156 by having the clients send a start time on each message, then using `actual start time = min(received reco start times)`

PR #158, opened 2023-03-24
Predictive Scanning & Variable N-Sides [minor]

### LINKED ISSUE #79: Configurable N-Side Refinement Iterations

### LINKED ISSUE #84: Iterations not fully asynchronous
The current handling of each iteration isn’t fully asynchronous, as the skymap_scanner waits for the previous nside to complete before starting the next one. The existing followup scanner bot will start up the next iteration once 30% have been scanned, which probably gives a nice speed up as some pixels can take much longer than others.

Related to #79

ric-evans:
# Predicitve Scanning
closes #84 
This will start the next round of nside pixels *before* all the recos for that nside (and previous nsides) have been collected. Note, when waiting for the final nside's recos, the threshold is ignored.

The threshold is a number `[0.1, 1.0]` available as `--predictive-scanning-threshold`, with a default of `1.0`.

The threshold is applied to the latest nside's reco'd pixels (not the percentage of overall collected recos).

# Variable N-Side Progression
closes #79 
Uses a dictionary of nsides & pixel extension numbers for generating subsequent pixels
- example: `--nsides 8:12 64:12 512:24` (first pixel-extension number must be 12)

closes https://github.com/icecube/skymap_scanner/issues/155

# Data Type Changes / Refactoring / README Updates
see https://github.com/icecube/skymap_scanner/blob/async-iterations/README.md#data-types

# Misc

closes https://github.com/icecube/skymap_scanner/issues/94


ric-evans:
@tianluyuan I snuck in a follow-up fix from #157, so feel free to ignore anything that deals with timestamps/runtimes

skymap_scanner/server/start_scan.py
```diff
@@ -419,30 +466,54 @@ async def collect(
...
+        await self.reporter.record_pixreco(
+            pixreco.nside,
+            pixreco_runtime,
+            roundtrip_start=self._pixreco_ids_sent__with_times[pixreco.id_tuple],
+            roundtrip_end=time.time(),
+        )
+
+    def is_scan_done(self) -> bool:
+        """Has every pixel been collected?"""
+        return self.n_received == self.n_sent
```
tianluyuan, line 485: This is a change from the previous `if col.pixreco_ids_sent == col.pixreco_ids_received`. Does checking the lengths here guarantee the received set of pixrecos is equal to the sent? Perhaps better to compare the keys of the sent pixel dictionary to the received set?

skymap_scanner/server/start_scan.py
```diff
@@ -542,40 +597,71 @@ async def serve_scan_iteration(
...
+    from_clients_queue: mq.Queue,
+    reco_algo: str,
+    nsides_dict: NSidesDict,
+    pixeler: PixelsToReco,
+    reporter: Reporter,
+    predictive_scanning_threshold: float,
+) -> int:
+    """Run the next (or first) scan iteration (set of pixel-recos).
+
+    Return the number of pixels sent. Stop when this is 0.
```
tianluyuan, line 617: `Stop when this is 0.` should be `Stop when all sent pixels have been received`?

skymap_scanner/server/choose_new_pixels_to_scan.py
tianluyuan, line 1: On L124 there is a factor of 0.3 to start the iterative refinement when at least 30% of the current iteration have been received. How does this interplay with the `predictive_scanning_threshold`?

tianluyuan (commented):
I think this is looking good and should speed up things I think. How do the skydriver tests look? I can try to test manually as well.

tests/data/reco_pixel_pkls/BRONZE/1660761104.474899.out.pkl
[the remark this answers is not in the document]
ric-evans: There is, I've attached some metadata to the message object for tracking client runtime

skymap_scanner/server/start_scan.py
```diff
@@ -419,30 +466,54 @@ async def collect(
...
+        """Has every pixel been collected?"""
+        return self.n_received == self.n_sent
 
+    def ok_to_serve_more(self) -> bool:
+        """Return whether enough pixel-recos collected to serve more.
+
+        If we are approaching the end of the scan, always return False.
+        """
+        if self._end_game:
+            return False
```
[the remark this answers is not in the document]
ric-evans: That's already implemented. The `Reporter` send a non-final ScanResult to SkyDriver (as opposed to saving to disk). From there, a SkyDriver user can build their plot, see https://github.com/icecube/skymap_scanner/blob/async-iterations/resources/utils/plot_skydriver_scan_result.py (this can move to https://github.com/icecube/[skyreader](https://github.com/icecube/skyreader), once it's set up)

skymap_scanner/server/start_scan.py
ric-evans: Good point, I've added back the check

skymap_scanner/server/choose_new_pixels_to_scan.py
ric-evans: I removed the hard-coded threshold & made a minimum of `0.1` at the command-line level

ric-evans:
> I think this is looking good and should speed up things I think. How do the skydriver tests look? I can try to test manually as well.

I'll test this in skydriver with a branch image before I merge

skymap_scanner/server/start_scan.py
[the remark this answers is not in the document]
tianluyuan: Ok great! In that case perhaps it'd be worth plotting a few intermediate results from a full iterative scan when testing this PR with SkyDriver, so we can see how the progression looks.

skymap_scanner/server/start_scan.py
```diff
@@ -162,31 +186,31 @@ def refine_vertex_time(vertex, time, direction, pulses, omgeo):
...
+    def gen_new_pixel_pframes(
+        self,
+        already_sent_pixels: Set[SentPixel],
+    ) -> Iterator[icetray.I3Frame]:
+        """Yield pixels (PFrames) to be reco'd."""
+
+        def pixel_already_sent(pixel: Tuple[icetray.I3Int, icetray.I3Int]) -> bool:
+            # Has this pixel already been sent? Ignore the position-variation id
+            for sent_pixel in already_sent_pixels:
+                if sent_pixel.nside == pixel[0] and sent_pixel.pixel_id == pixel[1]:
```
tianluyuan, line 198: How would it handle a scenario where we have N>1 position variations, and each pixel has finished reconstructing only one positional variation each so that it crosses the threshold for refinment, but no best-pixelreco has been recorded since none of the pixels are complete?

skymap_scanner/server/start_scan.py
```diff
@@ -419,30 +499,78 @@ async def collect(
...
+            return False
+
+        # get percentage of latest nside (otherwise BIG + small >> threshold)
+        latest_saved_nside = max(self.nsides_dict.keys())
+
+        # Have we already said yes to this nside?
+        # Ex: nside=8 reached threshold so we sent out nside=64,
+        #     now we're asking about nside=8 again--it's not okay to serve more.
+        if latest_saved_nside < max(self._sent_pixels_by_nside.keys()):
+            return False
```
tianluyuan, line 548: I'm not so sure this should be `False`. Since we start the next iteration before the current one is complete, it may be that there is a better reco found after the threshold was first crossed. So I think we'd want to run the refinement code each time a new pixel is received, even if it is for a lower nside. Does that make sense?

tianluyuan (commented):
I left some comments inline. After some thought I'm not sure if it's better to ignore positional-variations, or define the threshold on fully completed pixels after all positional-variations have been received back. I think the latter is probably better, so that the threshold is defined per pixel, not per (pixel, posvar).

skymap_scanner/config.py
```diff
@@ -16,8 +15,13 @@
 
 DEFAULT_GCD_DIR: Path = Path("/opt/i3-data/baseline_gcds")
 
-MIN_NSIDE_DEFAULT: Final = 8
-MAX_NSIDE_DEFAULT: Final = 512
+NSideProgression = Tuple[Tuple[int, int], ...]
+FIRST_NSIDE_PIXEL_EXTENSION = 12  # this is mandated by HEALPix algorithm
```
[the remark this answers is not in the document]
ric-evans: It's not needed. I figure we need a value there so it works with the dict-like nature of the nside-progression object & also the CL arg. I also figured it'd be better to keep this constant. Would `0` be better?

skymap_scanner/server/start_scan.py
ric-evans: the threshold for refinement is calculated by # of finished whole pixels (that's what gets stored in `nsides_dict`). However I could add an additional filter check at the `yield` in `gen_new_pixel_pframes()`. What do you think?

tianluyuan: I think it makes sense now that it treats it using the # finished pixels. What do you mean by additional filter check to look at the variation id?

skymap_scanner/server/start_scan.py
```diff
@@ -386,19 +389,38 @@ def __init__(
...
+        # percentage progress trackers
+        self._thresholds = PixelRecoCollector._make_thresholds(
+            predictive_scanning_threshold
+        )
+        self._nsides_percents_done = {n: 0.0 for n in nsides}
+
+    @staticmethod
+    def _make_thresholds(predictive_scanning_threshold: float) -> List[float]:
+        """Ex: predictive_scanning_threshold=0.66 -> [0.66, .7, 0.8, 0.9, 1.0]."""
         if not (0.0 < predictive_scanning_threshold <= 1.0):
```
tianluyuan, line 413: Should this minimum be 0.1?

tianluyuan (commented):
The thresholds seem reasonable. Let's see how the test does on SkyDriver.

skymap_scanner/server/start_scan.py
ric-evans: I'm not sure, I thnk I was hungry :sweat_smile: Sounds like we're good on this one

skymap_scanner/server/start_scan.py
ric-evans: right, good catch

skymap_scanner/server/start_scan.py
```diff
@@ -692,7 +729,20 @@ async def _serve_and_collect(
...
+                # we want to open re-refinement for all nsides <= max_nside_to_refine
+                nside_progression.get_slice_plus_one(max_nside_to_refine),
+            )
+            # Check if scan is done --
+            # we collected everything & there was no re-refinement of a region
+            if potentiallly_done:
+                if not sent_pixels:
+                    LOGGER.info("Done receiving/saving pixel-recos from clients.")
+                    return collector.n_sent
+                potentiallly_done = False
```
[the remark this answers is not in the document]
ric-evans: Say we have 8-64-512 nsides. 8 is done. We refined 64's pixels before it finished (that's normal). Now all of 512 is done, but we're still waiting on 64 to finish (slow nodes). Once 64 finishes we have technically received everything we sent, but we don't know if 64's newly received pixels warrant a re-refinement. So, we need to say we're potentially done and check again.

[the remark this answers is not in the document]
tianluyuan: Oh that makes sense. Good catch.

skymap_scanner/server/collector.py
```diff
@@ -0,0 +1,303 @@
...
+            raise ValueError(
+                f"`predictive_scanning_threshold` must be "
+                f"[{cfg.PREDICTIVE_SCANNING_THRESHOLD_MIN}, "
+                f"{cfg.PREDICTIVE_SCANNING_THRESHOLD_MAX}]: "
+                f"'{predictive_scanning_threshold}'"
+            )
+
+        thresholds = [predictive_scanning_threshold]
+        base = sorted([0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0])
+        thresholds.extend(base[bisect(base, predictive_scanning_threshold) :])
```
tianluyuan, line 133: Would it be possible to set the threshold to 1 for the last iteration NSide?

skymap_scanner/server/utils.py
```diff
@@ -40,3 +42,114 @@ def fetch_event_contents(
...
+                f"Invalid NSide Progression: the first pixel extension number must be {NSideProgression.FIRST_NSIDE_PIXEL_EXTENSION} ({int_int_list})"
+            )
+        if any(not isinstance(n[1], int) or n[1] <= 0 for n in int_int_list[1:]):
+            # don't check first extension #
+            raise ValueError(
+                f"Invalid NSide Progression: extension number must be positive int ({int_int_list})"
+            )
+        if any(not _is_pow_of_two(n[0]) for n in int_int_list):
+            raise ValueError(
+                f"Invalid NSide Progression: nside value must be positive n^2 ({int_int_list})"
```
tianluyuan, line 123: Should this be 2^n?

tianluyuan (commented):
Looks good to me! Maybe we should test two cases before merging, one with the `threshold=1` and another with `threshold=0.3` and see if the result looks similar.

skymap_scanner/server/utils.py
ric-evans: nice catch

ric-evans:
> Looks good to me! Maybe we should test two cases before merging, one with the `threshold=1` and another with `threshold=0.3` and see if the result looks similar.

Sure. I'll run those ASAP

PR #163, opened 2023-03-31
Add SplineMPE reconstruction [minor]

mlincett:
Tentative implementation of SplineMPE. I have partially imported the configuration of the scanning tool of @G-Sommani in the form of a I3Tray.


tianluyuan:
Great to have this in the scanner now. Have you been able to run any full-sky scans? Would be great to see how the contours look.

mlincett:
> Great to have this in the scanner now. Have you been able to run any full-sky scans? Would be great to see how the contours look.

Yes, I tested a couple of events. Contours shouldn't be much different from what we've seen from Giacomo, although the mechanics is slightly different and I do not apply the KS test.

I plan to test some more full sky scans now that the other major PRs have gone through.

tests/compare_scan_results.py
```diff
@@ -8,7 +8,7 @@
 from skyreader import SkyScanResult
 from wipac_dev_tools import logging_tools
 
-RTOL_PER_FIELD = {"llh": 0.18, "E_in": 0.7, "E_tot": 0.7}
+RTOL_PER_FIELD = {"llh": 0.22, "E_in": 0.7, "E_tot": 0.7}
```
ric-evans, line 11: what are your thoughts on using different rtols for each reco-algo?

ric-evans:
Also, don't forget to add `[minor]` to the PR title

tests/compare_scan_results.py
mlincett: It could be useful, although I am not sure this is really what we need.

Ideally we should be able to produce more "accurate" test data and have reasonably low tolerances, because with ~20% (~70% for energy reco) it's very easy to miss small changes.

I am afraid coarse scans are going to have bad/unstable likelihood values nevertheless, maybe we should "fabricate" test data according to the CI output instead of relying on Condor runs.

mlincett: By the way, we could get the test data out of CI "manually" by writing its JSON representation to stdout as part of the compare stage and using JSON for the reference files.

dsschult: Would it be possible to create a synthetic event that is fairly stable for coarse scans?  (or find one in real data, but that's by chance)  Then you could tighten down the tolerances, as it should always produce basically the same result.

mlincett: I have added a `cat` to print out the output JSON in CI for "realistic" scans. This can be used to extract the outputs from CI and replace the test data + tighten the tolerances.

Since now the compare stage is hardcoded to use `npz` I suggest we proceed with this change for all three recos after this PR is merged.

mlincett: Discuss this in #208.

tests/data/results_npz/splinempe/run00136766.evt000007637140.neutrino_1.json
```diff
@@ -0,0 +1,93 @@
...
+            "llh",
+            "E_in",
+            "E_tot"
+        ],
+        "data": [
+            [
+                0,
+                1899.893825371186,
+                NaN,
+                NaN
```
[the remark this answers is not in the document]
ric-evans: Does the CI test read this JSON file? There looks to be a similarly named npz file. That said, I do like the json format more for tests since it's human-readable

ric-evans (approved):

tests/data/results_npz/splinempe/run00136766.evt000007637140.neutrino_1.json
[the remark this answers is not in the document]
mlincett: > Does the CI test read this JSON file? There looks to be a similarly named npz file. That said, I do like the json format more for tests since it's human-readable

Not yet, but adding the JSON goes in that direction. My idea was to change the CI test to read JSON for all three recos once SplineMPE is merged.

I suppose `NaN` values can be handled gracefully even if they are not standard JSON. If not, I guess we can output zeros instead?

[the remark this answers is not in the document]
mlincett: OK, I am taking care of this non-destructively in this PR.

tests/compare_scan_results.py
mlincett: This is now fixed with updated test data.

.github/workflows/tests.yml
```diff
@@ -337,16 +337,21 @@ jobs:
...
           hese_event_01.json,
           run00136766-evt000007637140-GOLD.pkl,
           run00136662-evt000035405932-BRONZE.pkl
         ]
-        reco_algo: [
-          millipede_original,
-          millipede_wilks,
-          # splinempe
-        ]
+        exclude:
```
tianluyuan, line 350: Would it be cleaner to just have it run on hese as part of the test?

tianluyuan (approved):
Looks good to me.

.github/workflows/tests.yml
mlincett: Yes. At the beginning I was using `OnlineL2_SplineMPE` as seed so it was not available for HESE events. Now I take `VHESelfVeto` so we could run on HESE. Will try to get this in before merging.

mlincett: OK, this is not possible straight away because `OnlineL2_SplineMPE` is still used as energy reco seed. I am going to move forward with the merge for now but if a solution comes up I am happy to follow up in a future PR.

PR #168, opened 2023-04-03
Configuration parameter for pulse series name

mlincett:
There are many hardcoded occurrences of `pulsesName = SplitUncleanedInIcePulses`, i.e. the name identifying the pulse series used in the reconstruction.

Note that `prepare_frames()` used to support the passing of a different `pulsesName` but in a bad way: the user-selected pulse series would be used to overwrite the data of `SplitUncleanedInIcePulses`, so to match all the hardcoded values downstream.

With this PR:
- the value of `pulsesName` is specified in config.py as `cfg.INPUT_PULSES_NAME`;
- `pulsesName` becomes an optional argument of `extract_json_message()` defaulting to the value specified in `config`;
- defaults are removed from all the functions called by `extract_json_message()`;
- `pepare_frames()` no longer supports overriding/overwriting the default pulse series.

PR is draft for now ~~, as I am evaluating the opportunity of making the pulse series dependent on `reco_algo`. (Also, need to fix failing checks.)~~

Comments are welcome.

UPDATE: at this moment, it makes sense to make Skymap Scanner stick to `SplitUncleanedInIcePulses` as an input, and take care of pulse cleaning as part of the frame preparation / reco segments.


skymap_scanner/config.py
```diff
@@ -17,6 +17,7 @@
 
 
 # physics strings
+INPUT_PULSES_NAME: Final = "SplitUncleanedInIcePulses"
 INPUT_TIME_NAME: Final = "HESE_VHESelfVetoVertexTime"
 INPUT_POS_NAME: Final = "HESE_VHESelfVetoVertexPos"
 OUTPUT_PARTICLE_NAME: Final = "MillipedeSeedParticle"
```
tianluyuan, line 23: At some point it might make sense to move these configurations into the reco-specific modules.

tianluyuan (commented):
Looks good to me. Maybe we should open an issue to move some of the reco-specific input/output names into the reco-specific modules?

tianluyuan (approved):

ric-evans:
> defaults are removed from all the functions

skymap_scanner/config.py
ric-evans: This is a great idea

skymap_scanner/utils/extract_json_message.py
```diff
@@ -54,6 +54,7 @@ def extract_json_message(
     is_real_event: bool,
     cache_dir="./cache/",
     GCD_dir=None,
+    pulsesName = cfg.INPUT_PULSES_NAME
```
ric-evans, line 57: Do we need to make any of these optional? Can we just set the "standard" values within the function?

skymap_scanner/utils/extract_json_message.py
```diff
@@ -100,9 +102,9 @@ def __extract_frame_packet(
     frame_packet,
     reco_algo: str,
     is_real_event: bool,
+    pulsesName,
     cache_dir="./cache/",
     GCD_dir : str = str(cfg.DEFAULT_GCD_DIR),
-    pulsesName="SplitUncleanedInIcePulses",
```
ric-evans, line 105: see my comment above

skymap_scanner/utils/prepare_frames.py
```diff
@@ -77,9 +79,7 @@ def prepare_frames(frame_array, baseline_GCD, reco_algo, pulsesName="SplitUnclea
         simclasses,
     )
     
-    nominalPulsesName = "SplitUncleanedInIcePulses"
-    
-    output_frames = []
+    output_frames : list[icetray.I3Frame] = []
```
ric-evans, line 82: This typehint (subscripting the built-in `list`) is only supported in python 3.11. I suppose that's fine since we're confining the scanner to an image

ric-evans (changes requested):
a few things. Do we want to move to Python 3.11 as the minimum version? I'd support that decision

skymap_scanner/utils/extract_json_message.py
mlincett: No, indeed, better to make everything explicit in the call by `start_scan()`.

skymap_scanner/utils/prepare_frames.py
mlincett: Ah, right. For now I think it's better to stick with the < 3.11 style.

mlincett:
> a few things. Do we want to move to Python 3.11 as the minimum version? I'd support that decision

I would agree, but when I tried to install the package in a Python 3.11 environment some WIPAC dependencies were not available. I would stick to 3.10 style for this PR and see about moving to 3.11 in a new issue.

skymap_scanner/config.py
mlincett: Follow up in #173

skymap_scanner/utils/prepare_frames.py
mlincett: For 3.11 follow up in #172

skymap_scanner/utils/extract_json_message.py
```diff
@@ -100,9 +102,9 @@ def __extract_frame_packet(
     frame_packet,
     reco_algo: str,
     is_real_event: bool,
-    cache_dir="./cache/",
-    GCD_dir : str = str(cfg.DEFAULT_GCD_DIR),
-    pulsesName="SplitUncleanedInIcePulses",
+    pulsesName : str,
+    cache_dir : str,
+    GCD_dir : str,
```
ric-evans, line 107: Please excuse the pedantry for a moment. For pep8-compliance, there should be no space before `:`, like `pulsesName: str` instead of `pulsesName : str`. Looks like there are several places with this.

ric-evans (changes requested):
Other than the PEP8 thing, it looks good

skymap_scanner/utils/extract_json_message.py
mlincett: No worries, but how come black 23.1 is not enforcing this?

mlincett: Not even `autopep8` (2.0.2, `pycodestyle` 2.10.0) suggests changes. I took care of it manually.

ric-evans: `# fmt: off` is at the top of the file. I put that there to limit needless diffs

ric-evans (approved):

PR #170, opened 2023-04-04
Reconstruction-specific test files

mlincett:
Split test files in directories named after `reco_algo` and defines an appropriate test `matrix` for the workflows.


ric-evans (approved):
Good idea :+1:

PR #171, opened 2023-04-04
Perform scan in actual equatorial coordinates [minor]

### LINKED ISSUE #159: Scans currently performed in non-equatorial coordinates
Currently the scans are performed in a flipped equatorial coordinate system about the equator. This then is accounted for in the plotting/contour calculation [fixes](https://github.com/icecube/skymap_scanner/blob/863f7ce9f413818ad7b6b2edbdafc9ecde402413/skymap_scanner/utils/scan_result.py#L1182) are needed to produce a fits file in actual equatorial coordinates. It would be simpler and less confusing to perform the scan directly in equatorial coordinates.

https://github.com/icecube/skymap_scanner/blob/863f7ce9f413818ad7b6b2edbdafc9ecde402413/skymap_scanner/server/start_scan.py#L221
I think the above should just be `dec = numpy.pi/2.-dec`, in addition to similar updates in the plotting routines.

tianluyuan:
The original setup performed scans in flipped equatorial coordinates. Closes #159


tianluyuan:
I have verified that the scan is now done in equatorial coordinates and the plotting functions have been adapted to work for that new orientation.
Here is the before and after
Before             |  After
:-------------------------:|:-------------------------:
Screenshot 2023-04-04 at 16 47 44 | Screenshot 2023-04-04 at 16 47 24

For this event the new equatorial scan seems to recover some refined pixel in the lower-left of the llh space.

tianluyuan:
Oh those are screenshots, but you should be able to make a png by just changing the extension when calling savefig in scan_result.

ric-evans (approved):

PR #174, opened 2023-04-07
Fetch spline data from remote source using IceTray filestaging

mlincett:


dsschult (commented):
This looks pretty nice, and is something I've wanted for a while.

Q: Is there a reason to have a two-step process, where we need to call `stage_files()` first before `get_filename()`?  I would have thought they could be combined, with the staging happening as-needed.

The one case where I could see `stage_files()` being useful is when running multiple scanners in parallel, and calling that outside the algorithm on startup.

mlincett:
> This looks pretty nice, and is something I've wanted for a while.
> 
> Q: Is there a reason to have a two-step process, where we need to call `stage_files()` first before `get_filename()`? I would have thought they could be combined, with the staging happening as-needed.
> 
> The one case where I could see `stage_files()` being useful is when running multiple scanners in parallel, and calling that outside the algorithm on startup.

Thanks for the feedback @dsschult 

There is no strong practical reason for this, but conceptually:
- I tend to dislike patterns where the behavior of a method changes between one call and another because the method itself changes some (hidden) state;
- I prefer keeping read- methods and write- methods separate, and in particular I tend to dislike implicit write operations.

But if you think it is unnecessarily verbose I can merge the two.

dsschult:
> > This looks pretty nice, and is something I've wanted for a while.
> > Q: Is there a reason to have a two-step process, where we need to call `stage_files()` first before `get_filename()`? I would have thought they could be combined, with the staging happening as-needed.
> > The one case where I could see `stage_files()` being useful is when running multiple scanners in parallel, and calling that outside the algorithm on startup.
> 
> Thanks for the feedback @dsschult
> 
> There is no strong practical reason for this, but conceptually:
> 
>     * I tend to dislike patterns where the behavior of a method changes between one call and another because the method itself changes some (hidden) state;
> 
>     * I prefer keeping read- methods and write- methods separate, and in particular I tend to dislike implicit write operations.
> 
> 
> But if you think it is unnecessarily verbose I can merge the two.

No, it's fine the way it is.  Just wanted to hear your reasoning.

skymap_scanner/recos/millipede_wilks.py
```diff
@@ -44,16 +49,21 @@ class MillipedeWilks(RecoInterface):
...
+    effd_spline = datastager.get_filepath(FTP_EFFD_SPLINE)
 
     cascade_service = photonics_service.I3PhotoSplineService(
-        _base % "abs", _base % "prob", timingSigma=0.0,
-        effectivedistancetable = _effd,
+        abs_spline, prob_spline, timingSigma=0.0,
+        effectivedistancetable = effd_spline,
         tiltTableDir = os.path.expandvars('$I3_BUILD/ice-models/resources/models/ICEMODEL/spice_ftp-v1/'),
         quantileEpsilon=1
         )
```
ric-evans, line 69: Since this is much more dynamic now, it may be worth putting this into an `__init__()`. This way there are no surprises.

ric-evans (changes requested):
Generally, look good, but I do have some requests for coding style.

skymap_scanner/recos/millipede_wilks.py
ric-evans: Same goes for the other reco classes

dsschult: I think this was done to only load the tables once.  There may be a better way to do so (a singleton pattern?).

mlincett: @ric-evans I thought about that, but this would be a more significant change from the current design where everything in `RecoInterface` is static. I am just not sure whether the `traysegment` works when not defined as a static method.

skymap_scanner/utils/data_handling.py
```diff
@@ -0,0 +1,103 @@
...
+            subdir = source / self.local_subdir
+            filepath = subdir / basename
+            LOGGER.debug(f"Trying to read {filepath}...")
+            if filepath.is_file():
+                LOGGER.debug(f"-> success.")
+                filename = str(filepath)
+                break
+            else:
+                LOGGER.debug(f"-> fail.")
+        return filename
```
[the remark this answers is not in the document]
mlincett: Yes. This is a bad old habit of mine I need to overcome ;)

skymap_scanner/recos/millipede_wilks.py
ric-evans: Okay, those are decent reasons to keep it as is. It's not pretty, but it's functional. I would like to see how the shared memory pattern works with repeated calls now.

skymap_scanner/utils/data_handling.py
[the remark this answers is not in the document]
ric-evans: It's very C-like

skymap_scanner/recos/millipede_wilks.py
mlincett: I am also not a big fan of the current design where we do not instantiate the `RecoInterface` class, but I think it also makes a little sense to have the `traysegment` as a method. Maybe it can refactored so that the traysegment function is built and returned by a method of the class itself. But I would consider this for a future PR.

skymap_scanner/utils/data_handling.py
```diff
@@ -53,10 +54,24 @@ def stage_file(self, basename: str):
...
+            "-t",
+            "5",
+            "-O",
+            str(local_destination_path),
+            http_source_path,
+        ]
+        try:
+            subprocess.run(cmd, check=True)
+        except subprocess.CalledProcessError as e:
+            raise RuntimeError(f"Subprocess `wget` exited with status:\n -> {e}")
```
ric-evans, line 69: You can use the `from e` syntax in addition or instead of using the f-string. I'm not sure if the `CalledProcessError` instance already provides the command. You may be able to get away with not excepting at all

skymap_scanner/utils/data_handling.py
```diff
@@ -53,10 +54,24 @@ def stage_file(self, basename: str):
...
+
+        if not local_destination_path.is_file():
+            raise RuntimeError(
+                f"Subprocess `wget` succeeded but the resulting file is invalid:\n-> {cmd}"
+            )
 
     def get_filename(self, basename: str) -> str:
         """Look up basename under the local paths and the staging path and returns the first valid filename.
```
ric-evans, line 77: Does `basename` refer to the filename, similar to bash's `$(basename $FILE)`?

ric-evans (approved):
A couple of comments and questions, but no dealbreakers. Good work

skymap_scanner/utils/data_handling.py
mlincett: Yes. I call `basename` something like `bar.ext` and `filename` something like `/foo/bar.ext`.

Honestly I find "basename" not very intuitive. I would rather call `bar.ext` a `filename` and `/foo/bar.ext` a `filepath`, but then confusion may arise between paths in the form of `str` and paths in the form of `Path` objects... so I don't have a naming convention I am fully satisfied with right now.

ric-evans: > I would rather call bar.ext a filename and /foo/bar.ext a filepath

This is the convention I usually follow. If the variable name has "name" it's a string. If it has "path" it's either a string or a `Path`, but if it is a str, then it can be cast to a `Path` instance

skymap_scanner/config.py
```diff
@@ -15,6 +15,19 @@
 
 DEFAULT_GCD_DIR: Path = Path("/opt/i3-data/baseline_gcds")
 
+# Local data sources. These are assumed to be filesystem paths and are expected to have the same directory structure.
+LOCAL_DATA_SOURCES: Final[List[Path]] = [
+    Path("/opt/i3-data"),
+    Path("/cvmfs/icecube.opensciencegrid.org/data"),
+]
+# Directory path under a local data source to fetch spline data from.
+LOCAL_SPLINE_SUBDIR: Final[str] = "photon-tablex/splines"
```
tianluyuan, line 24: Is this a typo? Should be `photon-tables` right?

tianluyuan (approved):
LGTM as well. Thanks for implementing this. Could we run a test with the various cases
- not local but exists on cvmfs
- not on local or cvmfs but exists on remote
Just to make sure it's working as expected.

skymap_scanner/config.py
mlincett: Indeed. It was a dirty trick to force the CI test to stage the files. I will revert it back to the correct value.

mlincett:
> LGTM as well. Thanks for implementing this. Could we run a test with the various cases
> 
>     * not local but exists on cvmfs
> 
>     * not on local or cvmfs but exists on remote
>       Just to make sure it's working as expected.

We can CI-test the "not on local or cvmfs", I guess. For cvmfs I need to do it manually (I will try as part of the `splinempe-scan` branch dev).

dsschult:
> > LGTM as well. Thanks for implementing this. Could we run a test with the various cases
> > ```
> > * not local but exists on cvmfs
> > 
> > * not on local or cvmfs but exists on remote
> >   Just to make sure it's working as expected.
> > ```
> 
> We can CI-test the "not on local or cvmfs", I guess. For cvmfs I need to do it manually (I will try as part of the `splinempe-scan` branch dev).

If you really wanted to test the cvmfs version, you could create a special CI job with https://github.com/marketplace/actions/cernvm-file-system-cvmfs 

I assume the extra tests are only for the data_handling.py script, so you can pick files that you know are in various places.

skymap_scanner/config.py
ric-evans: If that's a useful test, you could use an env var (`ENV`)

mlincett:
Situation report:
- a test script / CI test has been added to test the staging (but without CVFMS, contributions on that are welcome);
- currently testing this on the field with the splinempe branch;
- will manually run a test for `millipede_wilks` before merging.

mlincett:
Situation report:
- a test script / CI test has been added to test the staging (but without CVFMS, contributions on that are welcome);
- currently testing this on the field with the splinempe branch;
- will manually run a test for millipede_wilks before merging.

PR #178, opened 2023-04-18
Migrate Files to `skyreader` & `skywriter` [minor]

### LINKED ISSUE #121: Migrate Non-Central Functionalities to `skyreader`/`skywriter`
A third-party repo/package would be useful for dealing with inputs & outputs, call it `skymap_scanner_tools`. This will have @tianluyuan's `i3_to_json.py`, `scan_result.py` (`ScanResult`), and `event_tools.py` (`EventMetadata`). The package would be used by the scanner, skydriver, and users.

In the SkyDriver POV, 
	- input is realtime's json format, and
	- output is `ScanResult`'s json format

Also, see:

>The current `.npz` is good as an internal output format but my idea was ultimately have conversions to portable data formats and plot generation taken care of by `ScanResult`.
> I hope I can take care of this more seriously starting next week.

_Originally posted by @mlincett in https://github.com/icecube/skymap_scanner/issues/9#issuecomment-1273488392_

and

>We now also have `ScanResult.read_json()` and `ScanResult.to_json()` for communication with SkyDriver (#109)
> Also, `scan_result.py` could be moved to a new `skymap_scanner_tools` repo for portability.
> I'm closing this issue since output formats are A. much better now and B. further improvements are outside the scope of the scanner proper

_Originally posted by @ric-evans in https://github.com/icecube/skymap_scanner/issues/9#issuecomment-1404244369_
            
and @tianluyuan's https://github.com/icecube/skymap_scanner/pull/83

ric-evans:
closes #121


ric-evans:
@tianluyuan we're using `metadata` wit the numpy `dtype`. Mypy is complaining about this--I add an in-line ignore since it seems to work. I did come across https://numpy.org/devdocs/reference/generated/numpy.dtype.metadata.html

requirements.txt
```diff
@@ -66,7 +69,7 @@ packaging==23.1
     #   astropy
     #   matplotlib
 pandas==2.0.0
```
tianluyuan, line 71: I wonder if some of these dependencies of `skyreader` could be made optional so that they don't get dragged into the skymap-scanner. For example `pandas` and `meander` are not used though `matplotlib` itself is a dependency of `healpy` so I guess that cannot be separated.

resources/utils/make_plots.sh
tianluyuan, line 1: This file seems useful but I couldn't find it in `skyreader`.

skymap_scanner/utils/to_skyscan_result.py
tianluyuan, line 1: Does it make sense to have this factory here or in `skyreader` itself?

tianluyuan (commented):
Looks good! Should this be a major version bump?

tianluyuan:
One thing that could be nice is to preserve commit history in the move. It's not so [difficult](https://stackoverflow.com/questions/1365541/how-to-move-some-files-from-one-git-repo-to-another-not-a-clone-preserving-hi) but not sure how you feel about going through the files.

requirements.txt
mlincett: I agree.

I guess this pattern is a consequence of `skyreader` providing both the data exchange format _and_
a library to analyze the results. One could think of a scenario where both `skyreader` and `skymap_scanner` depend on a third package that only defines the data format, but probably that would be excessive fragmentation.

ric-evans: That's right, it's a consequence of `skyreader`'s dependencies. I can make an install-extra for `skyreader` -- call it `plots`.

resources/utils/make_plots.sh
ric-evans: Sure I can include it in skyreader

mlincett (approved):
Looks OK to me as well!

requirements.txt
ric-evans: `pandas` is used by `serialzie()`. I looked into this further and came to the conclusion that it's not worth it: https://github.com/icecube/skyreader/pull/5#issuecomment-1515035778

resources/utils/make_plots.sh
ric-evans: Oh wait, the only reason this script was created was that icetray was required. Now that we don't need it, it's much easier to do this sort of thing. The `examples/` in skyreader have essentially this logic in them. Feel free to add your own scripts that you find useful :smile:

ric-evans: Also https://github.com/WIPACrepo/SkyDriver/tree/main/examples

tianluyuan (approved):

Issue #173, opened 2023-04-05
Single source of truth for reconstruction-specific logic


mlincett:
After the introduction of `millipede_wilks` and in view of the implementation of `splinempe` as `reco_algo`, we face several situations in which the code reads as the following pattern:
```
if reco_algo == "millipede_original":
    [...]
else:
    [...]
```
This makes particularly difficult to track down all the places in which a reco-specific logic is applied. In addition, different recos may require different inputs from the `p` frame (such as the vertex name). These are now centrally set in `config.py`.

It could be nice if all the reco specific logic could be provided in the reconstruction module (which by the way now is under `client`), similarly to what is done in the client for the reco I3Tray segment.

With the occasion we may also evaluate which steps should stay client-side and which should go server-side in `pepare_frames.py` (I am thinking of the pulse cleaning).

PR #189, opened 2023-05-09
Move all reco-specific logic to reco modules [minor]

mlincett:
Trying to address #173 paving the way for #163 

Built on top of #192 

All the `if reco_algo == [...]` clauses have been removed and now everything is sourced from the reco modules.


ric-evans:
I see this is still in Draft, so forgive my preemptive quasi-review...

In general, can you call `recos.get_reco_interface_object(...)` ASAP and then pass that object instead of the string? That way we're not referring to the same thing using two ways.

mlincett:
> I see this is in Draft, so forgive my preemptive quasi-review...
> 
> In general, can you call `recos.get_reco_interface_object(...)` ASAP and then pass that object instead of the string? That way we're not referring to the same thing using two ways.

No worry, feedback is always welcome. I am moving around some code and I was more worried about not breaking things than about optimizing the structure.

I agree with your suggestion. Also I intend get rid of the many occurrences of `reco_algo.lower()`.

skymap_scanner/recos/__init__.py
```diff
@@ -31,6 +37,40 @@ class RecoInterface:
...
+
+    @staticmethod
+    def do_rotate_vertex() -> bool:
+        """Defines whether each generated vertex variation should be rotated along the axis of the scan direction. With the exception for legacy algorithms (MillipedeOriginal) this should typycally return True."""
+        return True
+
+    @staticmethod
+    def do_refine_time() -> bool:
+        """Defines whether to refine seed time."""
+        return True
```
mlincett, line 56: Since these are pure booleans I am considering moving them to a dictionary with default values, from which individual recos can deviate with their own config in `__init__`.

Now we basically use these mostly/only to support the "legacy" configuration of `millipede_original`.

This would also make more straightforward to configure them at runtime.

mlincett: Dictionary is too free-form. Turned `RecoInterface` into an abstract class and declared annotated attributes.

tests/compare_scan_results.py
```diff
@@ -98,9 +98,9 @@ def compare_then_exit(
     )
 
     # increase tolerances
-    actual.require_close["llh"] = 2e-4
-    actual.require_close["E_in"] = 0.07
-    actual.require_close["E_tot"] = 0.07
+    actual.require_close["llh"] = 0.02
+    actual.require_close["E_in"] = 0.7
+    actual.require_close["E_tot"] = 0.7
```
ric-evans, line 103: A more robust pattern may be to change `is_close()` to include args for the various tolerances. That way we don't have magic numbers in the class. Plus it seems like the tolerances are variable for different recos and nsides, to a degree that `SkyScanResult` shouldn't try to be smart

mlincett: I agree. This was a quick fix to let the test pass but I agree that until we better understand how this variability can be reduced we should have a flexible `is_close()` definition.

skymap_scanner/server/start_scan.py
```diff
@@ -84,30 +86,14 @@ def __init__(
...
-                    dataclasses.I3Position(0., variation_distance,0.),
-                    dataclasses.I3Position(0.,0.,-variation_distance),
-                    dataclasses.I3Position(0.,0., variation_distance)
-                ]
-        else:
-            self.pos_variations = [
-                dataclasses.I3Position(0.,0.,0.),
-            ]
+        
+        self.reco_algo: str = reco_algo
```
ric-evans, line 90: Do we need to persist this separately from the reco algo object? We'd save some complexity by accessing the name via an instance variable of the object

skymap_scanner/server/start_scan.py
```diff
@@ -84,30 +86,14 @@ def __init__(
...
+        
+        RecoAlgo: type[RecoInterface] = recos.get_reco_interface_object(reco_algo)
+        
+        self.reco: RecoInterface = RecoAlgo()
```
ric-evans, line 94: Optional, but I'd combine this into one line since `RecoAlgo` isn't used after this line explicitly

skymap_scanner/server/start_scan.py
```diff
@@ -259,14 +267,22 @@ def _gen_pframes(
...
                     energy = self.nsides_dict[coarser_nside][coarser_pixel].energy
 
-        for i in range(0,len(self.pos_variations)):
+        # Now generate the vertex seed position variations according to the reco-specific logic.
+
+        n_pos_variations = len(self.pos_variations)
+
+        LOGGER.debug(f"Generating {n_pos_variations} position variations.")
+
+        for i in range(0, n_pos_variations):
```
ric-evans, line 276: python's `len()` is O(1), so you don't technically need `n_pos_variations`. While you're at it, `for i in range(0, n_pos_variations)` could be replaced with iterating `enumerate(self.pos_variations)`

skymap_scanner/utils/prepare_frames.py
```diff
@@ -67,18 +70,13 @@ def Process(self):
...
-        recclasses,
-        simclasses,
-    )
+def prepare_frames(frame_array, baseline_GCD: Union[None, str], reco_algo: str, pulsesName: str) -> List[icetray.I3Frame]: # type hint using list available from python 3.11
+
+    # ACTIVATE FOR DEBUG
+    # icetray.logging.console()
+
+    # Reconstruction algorithm provider class
+    RecoAlgo = recos.get_reco_interface_object(reco_algo)
```
ric-evans, line 79: Does this return an instance or a class? The case convention here implies a class, but the function implies an instance

tests/data/reco_pixel_pkls/get_toclient_msg_pkl.py
```diff
@@ -27,8 +27,16 @@ def main():
...
+    # When extracting the debug .pkl from ewms-pilot, the in- pickles already contain the full message.
+    # Do we need to support "bare" pframes pickles at all?
+    # For the moment, this is a workaround.
+    if "pframe" in pframe:
+        # Effectively this is equivalent to copying the file.
+        with open(args.pframe_pkl.parent / "in.pkl", "wb") as f:
+            pickle.dump(pframe, f)
+    else:
+        with open(args.pframe_pkl.parent / "in.pkl", "wb") as f:
+            pickle.dump({"pframe": pframe, "reco_algo": args.reco_algo}, f)
```
ric-evans, line 39: You're right. I believe the `milipede_original` test pickles are just pframes -- I don't remember how that came to be. But, I think it's fine since the data is okay. So we can leave the workaround. Can you add a comment explaining why?

ric-evans (changes requested):
You've made some great updates that'll help this all age gracefully -- take that tech debt!

I spotted some gotchas, but nothing major -- mostly style & documentation.

skymap_scanner/server/start_scan.py
mlincett: It would make sense but I would keep it this way for the sake of readability (especially if we add arguments to the constructor in the future).

skymap_scanner/server/start_scan.py
mlincett: I just didn't like writing `len()` many times. But I am happy to switch to `enumerate` and get rid of the auxiliary variable.

skymap_scanner/utils/prepare_frames.py
mlincett: Since we may want to add arguments to the RecoAlgo constructor in the future, this is supposed to be a class. I will rename `get_reco_interface_object` (it was returning a class from the beginning, after all).

tests/data/reco_pixel_pkls/get_toclient_msg_pkl.py
mlincett: See issue #196 for this. I'd say replacing thest pickles for `millipede_original` and making all the test data uniform should not cost a lot of time, and will make this simpler.

skymap_scanner/server/start_scan.py
mlincett: Possibly this is now unused, or it should be. Will check.

skymap_scanner/server/start_scan.py
ric-evans: Sounds good :+1:

tests/data/reco_pixel_pkls/get_toclient_msg_pkl.py
ric-evans: I hadn't seen #196. Those are all good points -- feel free to take those on. I'm indifferent whether those updates go into this PR or a separate one

skymap_scanner/recos/common/vertex_gen.py
```diff
@@ -0,0 +1,70 @@
...
+            dataclasses.I3Position(0.0, -radius, 0.0),
+            dataclasses.I3Position(0.0, radius, 0.0),
+            dataclasses.I3Position(0.0, 0.0, -radius),
+            dataclasses.I3Position(0.0, 0.0, radius),
+        ]
+
+    @staticmethod
+    def cylinder(
+        v_ax: List[float] = [-40.0, 40.0],
+        r_ax: List[float] = [150.0],
```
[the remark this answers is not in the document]
mlincett: Good point. I guess for now this can be a tuple.

skymap_scanner/recos/dummy.py
```diff
@@ -30,6 +30,7 @@ class Dummy(RecoInterface):
     """Logic for a dummy reco."""
 
     def __init__(self):
+        super().__init__
```
ric-evans, line 33: Maybe you're planning to do something with this, but is this a typo?

ric-evans:
I attached a couple issues, please change these if they are not covered by this PR, @mlincett

skymap_scanner/recos/dummy.py
mlincett: This was accidentally commited. Will remove.

ric-evans (approved):

mlincett:
@tianluyuan let me know if you also want to take a look before merging.

skymap_scanner/recos/common/pulse_proc.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def mask_deepcore(frame, origpulses: str, maskedpulses: str):
+    """Masks DeepCore pulses by selecting string numbers."""
+    FIRST_DEEPCORE_STRING: Final[int] = 79
+    frame[maskedpulses] = dataclasses.I3RecoPulseSeriesMapMask(
+        frame,
+        origpulses,
+        lambda omkey, index, pulse: omkey.string < FIRST_DEEPCORE_STRING,
+    )
```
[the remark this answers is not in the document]
mlincett: Almost done. For some reason with this change the CI test started failing on one of the pixels (reco loss reco). Now I've got very curious and I am trying to pinpoint exactly what triggers this behavior (at the cost of a large number of commits).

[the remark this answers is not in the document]
mlincett: Apparently it was just a default value I overlooked. All good.

skymap_scanner/recos/__init__.py
```diff
@@ -27,21 +30,25 @@ def __init__(self, reco_algo: str):
...
-class RecoInterface:
+class RecoInterface(ABC):
     """An abstract class encapsulating reco-specific logic."""
 
-    # List of spline file basenames required by the class.
+    name: str = __name__
+    # Reco-specific behaviors that need to be defined in derived classes.
+    rotate_vertex: bool
+    refine_time: bool
+    use_fallback_position: bool
```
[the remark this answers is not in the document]
mlincett: Changed to `add_fallback_position`.

Issue #191, opened 2023-05-11
Spline staging should be invoked only client-side


mlincett:
While developing #174 I have tried to keep the overall structure of the reconstruction classes as it was, where everything in the class is statically defined.

Now, in the current implementation, the staging of the spline tables occurs at the first import of the reconstruction module.

I generally dislike scenarios where "write" operations are executed upon import statements. At least the entire "staging of spline data" should be part of a static method rather than be defined in the class definition.

This is even more important since the reconstruction class is going to incorporate server-side logic (#189 ) and we do not want the server to download spline data when only the client needs it.

Issue #188, opened 2023-05-09
Runtime configuration of reconstruction algorithm


mlincett:
Presently, reconstruction algorithms have some hardcoded behaviors that we may want to configure at runtime. For example, the number of tested vertex positions.

In some cases, these can be tuned with environment variables (like the "mini" scan configuration, see `start_scan.py`).

We can think of a scenario in which we pass an additional string (JSON?) as config for the reconstruction algorithm.

Issue #148, opened 2023-03-04
CI Tests for Milipede Wilks


ric-evans:
It will be useful to have CI tests using `millipede_wilks` (nside=1). We may even save runtime.

tianluyuan:
I think the new cascade splines were removed from the Dockerfile to reduce space requirements. But there is the `no_cvmfs` Dockerfile, which I suppose could be used for CI? What are the constraints there?

mlincett:
@ric-evans @dsschult can we switch CI to use the `no_cvmfs` image or do we risk hitting some storage limit for GitHub workflows?

dsschult:
We might hit the storage limit.  I know for a whlie we were at the limit, before some cleanup of the images.

Also, larger images slow down CI by a noticeable amount.

mlincett:
> We might hit the storage limit. I know for a whlie we were at the limit, before some cleanup of the images.
> 
> Also, larger images slow down CI by a noticeable amount.

I understand. For CI this means we may need a different Dockerfile for each `reco_algo`, since each reco needs a different set of spline tables. This requires a heavy restructuring of the current workflows, I guess.

For reco-specific development branches it is still possible to amend the Dockerfile to fetch all the required data as long as the branch is being developed, and revert it back just before merging. It is not ideal but suffices for me right now.

mlincett:
> The other option is to download the spline tables at runtime in the "lite" image. Which wouldn't be much different size-wise than downloading/building that layer in the image, but is more customizable to the algorithm.

I guess we could begin with removing the spline tables from the IceTray image and fetching them in the Skymap Scanner image, see my remark in #166.

ric-evans:
I'm not a fan of keeping sizeable data inside the image--unless we need to version-track that data. If we can do an at-runtime pattern, I'd like that better

mlincett:
Thanks to #166 and the quick fix to #193 I think I can easily run a test scan and export the pickle files now. I am considering including this in #189 .

PR #192, opened 2023-05-11
Decouple spline staging from class declaration

mlincett:
Addressing #191 as intermediate step for #189.

This PR decouples the spline staging and instantation of the photospline services from the class declaration, so that importing the class does not trigger file writes or memory allocation by I3PhotoSplineService.

I have the kept the structure of having everything implemented in static attributes / methods, so the class is still not meant to be instantiated, but I am considering taking a step further and changing this as well.

In principle, aside from readability concerns, there is nothing preventing us to define the reconstruction traysegment as a bound method (so I have been told), so instead of:
```
tray.add(RecoAlgo.traysegment)
```
one would have:
```
reco = RecoAlgo(...)
tray.add(reco.traysegment, ...)
```
I think this would provide more flexibility as well as a natural way of implementing configurability for the reconstruction algorithm (see #188).

### DIFF
--- a/skymap_scanner/recos/__init__.py
+++ b/skymap_scanner/recos/__init__.py
@@ -31,6 +31,21 @@ class RecoInterface:
     # The spline files will be looked up in pre-defined local paths or fetched from a remote data store.
     SPLINE_REQUIREMENTS: List[str] = list()
 
+    def init(self):
+        raise NotImplementedError()
+
+    def setup_reco(self):
+        """Performs the necessary operations to prepare the execution of the reconstruction traysegment."""
+        raise NotImplementedError()
+
+    def get_datastager(self):
+        datastager = DataStager(
+            local_paths=cfg.LOCAL_DATA_SOURCES,
+            local_subdir=cfg.LOCAL_SPLINE_SUBDIR,
+            remote_path=f"{cfg.REMOTE_DATA_SOURCE}/{cfg.REMOTE_SPLINE_SUBDIR}",
+        )
+        return datastager
+
     @staticmethod
     def traysegment(tray, name, logger, **kwargs: Any) -> None:
         raise NotImplementedError()
@@ -49,7 +64,7 @@ def get_all_reco_algos() -> List[str]:
     ]
 
 
-def get_reco_interface_object(name: str) -> RecoInterface:
+def get_reco_interface_object(name: str) -> type[RecoInterface]:
     """Dynamically import the reco sub-module's class."""
     try:
         # Fetch module
--- a/skymap_scanner/recos/__init__.py
+++ b/skymap_scanner/recos/__init__.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING, Any, List
 if TYPE_CHECKING:  # https://stackoverflow.com/a/65265627
     from ..utils.pixel_classes import RecoPixelVariation
 
+import ..config as cfg
 from ..utils.data_handling import DataStager
 
 try:  # these are only used for typehints, so mock imports are fine
--- a/skymap_scanner/recos/__init__.py
+++ b/skymap_scanner/recos/__init__.py
@@ -8,7 +8,7 @@ from typing import TYPE_CHECKING, Any, List
 if TYPE_CHECKING:  # https://stackoverflow.com/a/65265627
     from ..utils.pixel_classes import RecoPixelVariation
 
-import ..config as cfg
+from .. import config as cfg
 from ..utils.data_handling import DataStager
 
 try:  # these are only used for typehints, so mock imports are fine
--- a/skymap_scanner/recos/__init__.py
+++ b/skymap_scanner/recos/__init__.py
@@ -8,6 +8,8 @@ from typing import TYPE_CHECKING, Any, List
 if TYPE_CHECKING:  # https://stackoverflow.com/a/65265627
     from ..utils.pixel_classes import RecoPixelVariation
 
+from ..utils.data_handling import DataStager
+
 try:  # these are only used for typehints, so mock imports are fine
     from icecube.dataclasses import I3Position  # type: ignore[import]
     from icecube.icetray import I3Frame  # type: ignore[import]
--- a/skymap_scanner/recos/__init__.py
+++ b/skymap_scanner/recos/__init__.py
@@ -54,8 +54,7 @@ def get_reco_interface_object(name: str) -> RecoInterface:
     try:
         # Fetch module
         module = importlib.import_module(f"{__name__}.{name.lower()}")
-        # Build the class name (i.e. reco_algo -> RecoAlgo).
-        return getattr(module, "".join(x.capitalize() for x in name.split("_")))
+        return module.RECO_CLASS
     except ModuleNotFoundError as e:
         if name not in get_all_reco_algos():
             # checking this in 'except' allows us to use 'from e'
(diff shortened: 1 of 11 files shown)


ric-evans:
> In general I'm in favor of not loading in data files until it's needed. I think this would cause issues for parallelization as in #140, though I'm not sure if that is still being pursued.
> 
> Perhaps @kjmeagher and @briedel can comment, or at least take note of this PR.

Ah yes, parallelization :smile:  `ewms-pilot` enabled on-worker parallelization (called "multi-tasking") in `v0.9.0` (https://github.com/Observation-Management-Service/ewms-pilot/pull/28). Now that https://github.com/icecube/skymap_scanner/pull/174 has been merged, this can be enabled.

kjmeagher:
This is fine with me. This is basically the same thing as I suggested

ric-evans:
new issue: https://github.com/icecube/skymap_scanner/issues/194

skymap_scanner/recos/millipede_original.py
```diff
@@ -159,9 +163,20 @@ def LatePulseCleaning(frame, Pulses, Residual=3e3*I3Units.ns):
...
+    @staticmethod
     @icetray.traysegment
     def traysegment(tray, name, logger, seed=None):
         """Perform MillipedeOriginal reco."""
+        datastager = MillipedeOriginal.init_datastager()
+
+        abs_spline = datastager.get_filepath(MillipedeOriginal.MIE_ABS_SPLINE)
+        prob_spline = datastager.get_filepath(MillipedeOriginal.MIE_PROB_SPLINE)   
+
+        cascade_service = photonics_service.I3PhotoSplineService(abs_spline, prob_spline, timingSigma=0.0)
```
tianluyuan, line 175: Maybe to clarify my previous comment, I'm reading this as creating a new `I3PhotoSplineService` each time `traysegment` is run, while #140 took advantage of the fact that the `cascade_service` was created once at instantiation and then shared across processes. Isn't this going to duplicate the memory under parallelization?

ric-evans: I believe our conclusion is that we're going to try to leverage `/dev/shm`. Mounting is platform dependent, but that's not the responsibility of the client code. Perhaps @dsschult can elaborate (https://icecube-spno.slack.com/archives/GU72UBBC3/p1681835217915089)?

mlincett: It should be no problem to support such a scenario if instantiate the reco class and we set up the photospline services as attributes "on demand". I would like to move away from the current "all static" paradigm anyway, so I will update this PR.

It is not clear to me whether `ewms-pilot` multitasking will allow to use shared memory somehow. My understanding is that now a subprocess(?) is spawned for every pixel to be reconstructed. But this is a different topic altogether.

dsschult: Yes, the idea was to use `/dev/shm` shared memory, since it's allocating it anyway.  The key will be to correctly handle the loading/cleanup so we don't get a deadlock situation.  Not sure if that needs to be included here, or wait until #194?

mlincett: With the last changes the reconstruction class is instantiated in `reco_icetray.py`. It provides a `setup_reco()` method that stages the spline data and initializes the photospline services within the object. The reco traysegment is now a bound method.

This should allow more flexibility and better control re to resource allocation. Cannot really foresee whether the design is compatible with `ewms-pilot` multiprocessing but we can re-evaluate if necessary.

ric-evans: Right, for #148, I'll test this in different environments (container, non-container, condor) and measure memory usage.

skymap_scanner/recos/__init__.py
```diff
@@ -49,13 +67,12 @@ def get_all_reco_algos() -> List[str]:
     ]
 
 
-def get_reco_interface_object(name: str) -> RecoInterface:
+def get_reco_interface_object(name: str) -> type[RecoInterface]:
```
ric-evans, line 70: I think this syntax is py3.11 only. You'll need to import `Type`

skymap_scanner/recos/dummy.py
```diff
@@ -69,3 +75,7 @@ def to_recopixelvariation(frame: I3Frame, geometry: I3Frame) -> RecoPixelVariati
             time=frame["Dummy_time"].value,
             energy=frame["Dummy_time"].value,
         )
+
+
+# Provide a standard alias for the reconstruction class provided by this module.
+RECO_CLASS: Final[type[RecoInterface]] = Dummy
```
ric-evans, line 81: Since you're getting rid of the static resources, you can move this to the class itself as a getter/property. Then that'd force all new reco classes to implement it.

skymap_scanner/recos/millipede_wilks.py
```diff
@@ -47,31 +45,28 @@ class MillipedeWilks(RecoInterface):
...
 
-    # At HESE energies, deposited light is dominated by the stochastic losses
-    # (muon part emits so little light in comparison)
-    # This is why we can use cascade tables
-    datastager = DataStager(
-        local_paths=cfg.LOCAL_DATA_SOURCES,
-        local_subdir=cfg.LOCAL_SPLINE_SUBDIR,
-        remote_path=f"{cfg.REMOTE_DATA_SOURCE}/{cfg.REMOTE_SPLINE_SUBDIR}",
-    )
+    def setup_reco(self):
```
ric-evans, line 51: Can you add a doc block indicating that this method is used to load expensive resources?

ric-evans (commented):
Good improvements. I have a couple requests, mainly for coding style.

skymap_scanner/recos/dummy.py
mlincett: > Since you're getting rid of the static resources, you can move this to the class itself as a getter/property. Then that'd force all new reco classes to implement it.

I am not sure I understand. The problem we are trying to solve is looking up the class name via the module name, so this has to be defined at the module level but outside the class.

skymap_scanner/recos/__init__.py
mlincett: According to docs, `builtins.type` is supported since 3.9 and `Type` is actually deprecated.

skymap_scanner/recos/dummy.py
mlincett: I acknowledge that this is not very elegant. We could consider a manually compiled lookup table in `__init__.py` but  I am not sure I like that more.

skymap_scanner/recos/millipede_wilks.py
mlincett: Will do in `RecoInterface`.

skymap_scanner/recos/millipede_original.py
```diff
@@ -44,25 +43,32 @@ class MillipedeOriginal(RecoInterface):
     # Constants ########################################################
     pulsesName = cfg.INPUT_PULSES_NAME
     pulsesName_cleaned = pulsesName+'LatePulseCleaned'
-    SPEScale = 0.99
+
+    SPEScale = 0.99 # DOM efficiency
```
tianluyuan, line 47: Would it make sense to move all these variables to `__init__` and make `def exclusions` a bound method?

mlincett: I agree on the general idea of getting rid of `ClassName.constant` references.

I would still leave proper "constants" at the class level (we can still access them through `self` in bound/class methods after all).

I would keep `exclusions()` a class method as long as its parameters do not need to be defined at runtime, but of course we can change this in the future.

tianluyuan (approved):

skymap_scanner/recos/__init__.py
ric-evans: nice!

PR #202, opened 2023-06-30
Cloud Followup Updates [minor]

### LINKED ISSUE #179: Incomplete scans should be flagged as such
Using 3.8.2, I noticed some scans of the same event did not closely agree with scans using previous versions. It turned out the scans were incomplete but saved to disk. See below for an example.

Screenshot 2023-04-23 at 14 06 53

I think the issue occurs [here](https://github.com/icecube/skymap_scanner/blob/ec3a2c09791884025b0c8736a208d54d16eb0fa9/skymap_scanner/server/start_scan.py#L484), where `_serve_and_collect` can return even when the scan is incomplete. Looking at the state of the output files though there is no flag that marks it as any different from completed scans. We could add a flag to the metadata to indicate this.

ric-evans:
Found some bugs in the cloud test runs, here are the fixes

A bulk of the fixes are in the depedencies

closes #179


skymap_scanner/server/utils.py
```diff
@@ -1,27 +1,53 @@
...
+    *that* important. BUT if `infinite_retry=True`, then keep
+    retrying indefinitely.
+    """
+    while True:
+        try:
+            return await pfunc()
+        except requests.exceptions.HTTPError:
+            if not infinite_retry:
+                return {}
+        time.sleep(60)
```
dsschult, line 35: This should be an async sleep

dsschult (commented):
I'm unsure how much I like the idea of an infinite retry, since it will keep going forever, and basically hang that execution context there.  Also, with async, you could potentially build up an infinite amount of these retries if they get run in parallel.  The server would not thank you for initiating a DDOS.

ric-evans:
> I'm unsure how much I like the idea of an infinite retry, since it will keep going forever, and basically hang that execution context there. Also, with async, you could potentially build up an infinite amount of these retries if they get run in parallel. The server would not thank you for initiating a DDOS.
> 
> Note that we already have a basic retry mechanism in place in the rest client (I think it tries up to 11 times?). In what cases is this not enough?

My idea is to essentially have two different timeouts. One for calls that are very important: getting the event and sending the result. All other calls matter much less. I ran into a situation where the REST server had a bug and it took ~20 mins to roll out a fix--which is pretty fast.

Could I instead increase the # of retries momentarily?

dsschult:
> > I'm unsure how much I like the idea of an infinite retry, since it will keep going forever, and basically hang that execution context there. Also, with async, you could potentially build up an infinite amount of these retries if they get run in parallel. The server would not thank you for initiating a DDOS.
> > Note that we already have a basic retry mechanism in place in the rest client (I think it tries up to 11 times?). In what cases is this not enough?
> 
> My idea is to essentially have two different timeouts/retries. One for calls that are very important: getting the event and sending the result. All other calls matter much less. I ran into a situation where the REST server had a bug and it took ~20 mins to roll out a fix--which is pretty fast.
> 
> Could I instead increase the # of retries momentarily?

You would have to reset the session object, but potentially yes.
https://github.com/WIPACrepo/rest-tools/blob/master/rest_tools/client/client.py#L69 

Note that this might kill any existing requests using that rest client.  It may be smoother to have a second rest client with a different timeout/retries setting.

We might also think of putting better logic in rest-tools if there's a case it doesn't handle well.  Maybe increasing the backoff for retries?

ric-evans:
TODO: add `transfer_output_files = ""` to client starter

skymap_scanner/server/start_scan.py
```diff
@@ -725,13 +732,15 @@ def _nside_and_pixelextension(val: str) -> Tuple[int, int]:
...
         timeout=cfg.ENV.SKYSCAN_MQ_TIMEOUT_TO_CLIENTS,
+        ack_timeout=cfg.SERVER_ACK_TIMEOUT,
     )
     from_clients_queue = mq.Queue(
         cfg.ENV.SKYSCAN_BROKER_CLIENT,
         address=cfg.ENV.SKYSCAN_BROKER_ADDRESS,
         name=f"from-clients-{scan_id}",
         auth_token=cfg.ENV.SKYSCAN_BROKER_AUTH,
         timeout=cfg.ENV.SKYSCAN_MQ_TIMEOUT_FROM_CLIENTS,
+        ack_timeout=cfg.SERVER_ACK_TIMEOUT,
```
ric-evans, line 743: We don't want this. The server should have short heartbeats to avoid TCP shenanigans. Call `connection.process_data_events()` occasionally (perhaps abstract away to mqclient). This issue we're trying to address is when the server has transitioned to the collection phase while the sending-connection is sitting idle

ric-evans:
The logs show the 1st (of 2) worker is retrieving _all_ the messages in its backlog, and so the 2nd worker quits early. This is because using pika's `consume` with a prefetch of `0` is essentially an **infinite** prefetch.

The solution will be to use pika's `basic.get` when the user wants a prefetch of `0`

ric-evans:
> > > I'm unsure how much I like the idea of an infinite retry, since it will keep going forever, and basically hang that execution context there. Also, with async, you could potentially build up an infinite amount of these retries if they get run in parallel. The server would not thank you for initiating a DDOS.
> > > Note that we already have a basic retry mechanism in place in the rest client (I think it tries up to 11 times?). In what cases is this not enough?
> > 
> > 
> > My idea is to essentially have two different timeouts/retries. One for calls that are very important: getting the event and sending the result. All other calls matter much less. I ran into a situation where the REST server had a bug and it took ~20 mins to roll out a fix--which is pretty fast.
> > Could I instead increase the # of retries momentarily?
> 
> You would have to reset the session object, but potentially yes. [WIPACrepo/rest-tools@`master`/rest_tools/client/client.py#L69](https://github.com/WIPACrepo/rest-tools/blob/master/rest_tools/client/client.py?rgh-link-date=2023-07-05T21%3A24%3A17Z#L69) ([ at 2023-07-05T21:24:17Z](https://github.com/WIPACrepo/rest-tools/blob/8c1be54386c72624ba40cafdc3c2d5062c5a029a/rest_tools/client/client.py?rgh-link-date=2023-07-05T21%3A24%3A17Z#L69))
> 
> Note that this might kill any existing requests using that rest client. It may be smoother to have a second rest client with a different timeout/retries setting.
> 
> We might also think of putting better logic in rest-tools if there's a case it doesn't handle well. Maybe increasing the backoff for retries?

https://github.com/WIPACrepo/rest-tools/issues/131

skymap_scanner/server/start_scan.py
ric-evans: Being fixed in https://github.com/Observation-Management-Service/ewms-pilot/pull/44

skymap_scanner/server/utils.py
```diff
@@ -8,24 +8,45 @@
...
+            # backoff_factor=0.3,
         )
-        LOGGER.info("Fetched event contents from SkyDriver")
-        return manifest["event_i3live_json_dict"]
+    else:
+        return RestClient(
+            cfg.ENV.SKYSCAN_SKYDRIVER_ADDRESS,
+            token=cfg.ENV.SKYSCAN_SKYDRIVER_AUTH,
+            timeout=10.0,
+            retries=0,
```
dsschult, line 32: Do we really want no retries at all?  This means it's OK for these requests to drop on the floor if there's a transient error.

dsschult (approved):

Issue #200, opened 2023-06-09
Instability of some numerical results (condor vs CI)


mlincett:
As we have been discussing for a while, it seems sometimes the numerical results of millipede are not stable across runs on different platforms, in spite of containerisation.

While I am not sure there is an "issue" to solve, I would like to track here some observations about this behaviour. Updates to come.

mlincett:
After #198 the test of `MillipedeWilks` fails with a pixel that fails the energy loss reco and has a difference in likelihood of 17%.

ric-evans:
this also seems relevant: https://github.com/icecube/skymap_scanner/issues/60

tianluyuan:
Could you clarify where the failure occurred? Naively I would have thought when run on GitHub CI the platform should be near identical, so tests should fail randomly. Though results from condor could still be different.

ric-evans:
> Could you clarify where the failure occurred? Naively I would have thought when run on GitHub CI the platform should be near identical, so tests should fail randomly. Though results from condor could still be different.

We're seeing it here https://github.com/icecube/skymap_scanner/pull/204 with millipede wilks

ric-evans:
Following up, the issue in https://github.com/icecube/skymap_scanner/pull/204 is not related to this after all

tianluyuan:
The spline tables for all these cases are obtained using `wget` over http. One possibility is that in those cases the integrity of the file is lost, leading to the different results. Perhaps we can add checksumming to ensure that is not happening.

tianluyuan:
I think the containers can be dataless and cvmfs-less if we really wanted, but in that case we should ensure the file transfer mechanism is robust. I don't think I have run into such issues on sub-2, but that could be because it's using the spline tables on cvmfs rather than over http.

mlincett:
I believe issues in downloads should result in truncated files, rather than corrupted data, but a checksum is definitely a good idea.

dsschult:
I've seen corrupted data without a truncation when transferring files.  It's a low probability to get a bit flip that passes the TCP checksum, but it does happen when moving around enough bytes.  Note that this does not happen with CVMFS, since that uses checksums internally.

The easiest thing to do is to have a checksum file next to the file you want to download, and if it's available then perform the checksum test.

tianluyuan:
[...]

PR #204, opened 2023-07-13
Write Results & Manifest to File Throughout Scan

ric-evans:
This helps when scans are being slow and are run manually


ric-evans:
> Any idea why some of the tests started failing? This seems some manifestation of #200 , although it's curious to see it happening for changes that do not touch the "core" reco steps, like the outcome depends on some "hidden" random state.

Your guess is ~as good as~ better than mine. At first glance, it looks like one pixel was different.

PR #206, opened 2023-07-18
Result Regression Error Fix

ric-evans:
The issue was the ftp-v1 timing spline table. @tianluyuan is working in this area and plans to have a fix. For now, let wilks tests pass until Sep 3, when the CI tests will fail (a la bash magic)

There is an unrelated bug with `ewms-pilot` / `oms-mqclient`--pin for now and fix in https://github.com/icecube/skymap_scanner/pull/202


requirements-all.txt
tianluyuan, line 1: did we want to rename these files?

tianluyuan (approved):

requirements-all.txt
ric-evans: done!

mlincett (approved):

PR #209, opened 2023-07-31
Use Script for CI Testing

ric-evans:
These scripts can be leveraged for generating CI test data locally

Note: The original use case for this PR is no longer relevant, so the commit history here may seem odd. Nonetheless, using a script for the test scans can help us avoid tech debt.


ric-evans:
Tests should pass after https://github.com/icecube/skyreader/issues/18

dsschult (approved):
Approved in concept, pending the tests passing.

PR #210, opened 2023-08-16
Bump to IceTray `v1.7.2`

ric-evans:
`v1.7.2` includes new updates to photonics-service


ric-evans:
Tested with `v1.7.0`. Waiting on `v1.7.2` to be released

tianluyuan:
@briedel or @blaufuss could we get a new base icetray image with 1.7.2?

PR #211, opened 2023-08-18
updates for timing

### LINKED ISSUE #208: Replace _realistic_ test data with JSON output from CI
In order to avoid keeping dangerously large tolerances, we can extract manually the JSON scan outputs from CI and provide them as test data.

tianluyuan:
Incorporate https://github.com/icecube/icetray/pull/3133 and other pulse timing updates

closes #208


ric-evans:
> The test updates here should be enough to address and close #208. Tolerances are tightened to 0.15, 0.1, and 0.1 as well.

can you add "closes #208" to the PR description?

tests/env-vars.sh
```diff
@@ -5,16 +5,16 @@ set -ex  # file is sourced so turn off at end
...
 export SKYSCAN_BROKER_CLIENT=rabbitmq
 # note=auth env vars are in job(s)
-export EWMS_PILOT_TASK_TIMEOUT=600
+export EWMS_PILOT_TASK_TIMEOUT=1800
 # export SKYSCAN_DEBUG_DIR=debug-pkl-dir -- rely on user value
 export SKYSCAN_MQ_TIMEOUT_TO_CLIENTS=5
-# export SKYSCAN_MQ_TIMEOUT_FROM_CLIENTS=60  # use default
-export SKYSCAN_MQ_CLIENT_TIMEOUT_WAIT_FOR_FIRST_MESSAGE=120
+export SKYSCAN_MQ_TIMEOUT_FROM_CLIENTS=1800
+# export SKYSCAN_MQ_CLIENT_TIMEOUT_WAIT_FOR_FIRST_MESSAGE=120
```
ric-evans, line 12: I'm fine with most of these env values. However, do we expect any reco to take 30 mins?

tianluyuan: hopefully not, but before it was using the default which was actually 3 days.

ric-evans (approved):

PR #227, opened 2023-10-14
debugging tests

tianluyuan:


tianluyuan:
It looks like the github actions runs on four (?) cpus. And tests fail for millipede_wilks when run on `Intel(R) Xeon(R) CPU E5-2*`. This is a processor that came out ca 2014 i think, and doesn't have avx512 from what lscpu is showing.

I'm not sure how the various avx support plays in with the dependencies we have and when running within a docker container. Perhaps for now we can just relax the tolerances when running on the older cpus, which is what i've set here. 

When tests are run on newer cpus, on which the reference data is based i imagine, the differences are at the 1e-12 level or lower.

ric-evans (approved):

tianluyuan:
Setting the following in the Dockerfiles to disable AVX512 seems to improve reproducibility. I do still observe some fluctuations with splinempe, which is why the llh tolerance is still set to `0.05`. It might be worth tracking down where those are coming from at some point.
```
ENV OPENBLAS_CORETYPE="Haswell"
ENV NPY_DISABLE_CPU_FEATURES="AVX512F,AVX512_KNL,AVX512_KNM,AVX512_CLX,AVX512_CNL,AVX512_ICL,AVX512CD,AVX512_SKX"
```

PR #244, opened 2023-11-30
Support realtime 2023a data format (`SplitInIcePulses`)

mlincett:
With these changes, ~~we set the name of the base pulse series to `SkyScanBasePulses`~~ we support the change in the input pulses name occurred with the realtime 2023a format. 

As part of the `prepare_frames`, a `pulses_name` argument is used as the source to copy the pulses from. When `pulses_name` is `None`, we use a map from the realtime JSON "version" to the pulses name ("SplitInIcePulses" for 2023a) and "SplitUncleanedInIcePulses" for 2021a. The latter is also the default.

In general, an arbitrary `pulses_name` is supported in `extra_json_message()`. In the future, this can be used to read an arbitrary series of pulses instead of the mandated one.

Out of simplicity, I have decided to copy over the pulses to a new "static" name rather than use it as an attribute of the reco object.


tianluyuan:
@mlincett I merged in the latest updates from main and updated the baseGCD url to point to `prod-exe`. Now that the Dockerfile is also pulling in the baseGCDs from prod-exe this might be redundant, but perhaps it's ok for now? Otherwise I think this is good to go and I'll go ahead and mark it as ready for review.

Was there anything else you were working on?

mlincett:
> @mlincett I merged in the latest updates from main and updated the baseGCD url to point to `prod-exe`. Now that the Dockerfile is also pulling in the baseGCDs from prod-exe this might be redundant, but perhaps it's ok for now? Otherwise I think this is good to go and I'll go ahead and mark it as ready for review.
> 
> Was there anything else you were working on?

Thank you for picking this up! I believe it should be good to go.

(We will not have single-pixel tests for the new format but I don't see that as strictly necessary for now.)

Initially, I meant to rename the input pulse series name of the scanner to something neutral and different from "SplitUncleanedInIcePulses", but this would require replacing all the single-pixel test data (or finding some workaround). I guess we can ignore that for now.

skymap_scanner/server/start_scan.py
```diff
@@ -691,16 +691,20 @@ def _nside_and_pixelextension(val: str) -> Tuple[int, int]:
...
+    # leave argument to support passing an arbitrary pulses_name in the future
+    LOGGER.info("Extracting event...")
     event_metadata, state_dict = extract_json_message.extract_json_message(
         event_contents,
         reco_algo=args.reco_algo,
         is_real_event=args.real_event,
         cache_dir=str(args.cache_dir),
         GCD_dir=str(args.gcd_dir),
-        pulsesName=cfg.INPUT_PULSES_NAME,
+        pulses_name=None,
```
ric-evans, line 703: When would this not be `None`? I'm concerned that this may be configuration creep, ie supporting more options than needed.

skymap_scanner/utils/extract_json_message.py
```diff
@@ -50,34 +51,60 @@ def extract_GCD_diff_base_filename(frame_packet):
 
     return GCD_diff_base_filename
 
+def __get_pulses_name(event_dict: dict):
```
ric-evans, line 54: Here and a few other new functions start with `__`. What's the reasoning to no use a single `_`? I think it's technically okay, but I believe python does some strange magic when class methods/attributes are prefixed with `__`.

skymap_scanner/utils/extract_json_message.py
```diff
@@ -229,36 +293,37 @@ def __extract_frame_packet(
...
-    else:
-        # frame_packet has either normal GCD or has been reassembled
-        frame_packet = prepare_frames(frame_packet, event_metadata, None, reco_algo, pulsesName)
+    # Uncompress GCD info and invoke `prepare_frames` traysegment provided by `reco_algo`
+    # - frame_packet has GCD diff => baseline_GCD_file is a path string
+    # - frame_packet has either normal GCD or has been reassembled => baseline_GCD_file is None
+    prepared_frame_packet = prepare_frames(frame_packet, event_metadata, baseline_GCD_file, reco_algo, pulses_name=pulses_name)
+
+    # Delete original frame packet.
+    del frame_packet
```
ric-evans, line 302: I'm not a fan of using `del` to "switch main characters", ie `frame_packet` is no longer relevant now that `prepared_frame_packet` exists. It's a sign that the function is becoming bloated. The function could be made much more readable with additional helper functions considering its current size. A refactor may be beyond the scope of this PR, perhaps open a new issue?

ric-evans (changes requested):
A few questions and requests

skymap_scanner/server/start_scan.py
mlincett: @tianluyuan at some point mentioned that it would be nice to be able to optionally take a different pulse series as input, on-demand. So either we establish a plan to support it, or we drop the argument.

Note that such a feature can be a bit tricky to implement, given that for a given pulse series, the reco algos may try to look up associated entries in the frame (for example the corresponding `TimeRange`).

skymap_scanner/utils/extract_json_message.py
mlincett: I think we use underscores here to mark a function that has been split out a previously existing function, due to the latter being too complicate. In this improvised naming convention  `__g` is a function that splits some code out of `_f` which was originally split out of some other function. Not sure whether I was the first to use the double underscore here or not.

However this is meant to be somewhat temporary, like: "I have split out some code that was previously there but probably we should review this more in depth". But I agree, we can avoid double underscores just to be safe.

skymap_scanner/utils/extract_json_message.py
mlincett: I agree! We can take a second look to see whether this can be divided into two functions.

skymap_scanner/server/start_scan.py
tianluyuan: Maybe we can just remove the argument for now. If that enhancement makes it in it can be added back.

skymap_scanner/utils/extract_json_message.py
tianluyuan: Now updated all double underscores to single underscore.

skymap_scanner/utils/extract_json_message.py
tianluyuan: Added to #260

skymap_scanner/utils/extract_json_message.py
ric-evans: > However this is meant to be somewhat temporary, like: "I have split out some code that was previously there but probably we should review this more in depth". But I agree, we can avoid double underscores just to be safe.

That explains it until the original intention is forgotten :smile: 

Tangent: I think there's nothing wrong with very long function names as long as it's accurately descriptive. But this _is_ a controversial opinion in the dev world

ric-evans (approved):

PR #258, opened 2023-12-22
[minor] Download baseline GCDs in Dockerfile and bump to icetray 1.9.1

### LINKED ISSUE #252: Update baseline GCDs
With 2023 run start we need to update the baseline GCDs. Baseline GCDs are fetched from `convey` in the generation of the base IceTray container. Hence the base IceTray container should be updated (*).

There seems to be no easy way for a Skymap Scanner developer to see:
- whether a new IceTray container has been produced;
- whether such container includes the required GCD (at least without trying to run the code once).

**Can we rebase Skymap Scanner on a newer IceTray including the latest GCD?**

(*) this workflow is far from optimal because the container content is determined by _when_ the container is generated rather than by the `Dockerfile` content.

Please note that as I have already observed in https://github.com/icecube/realtime/issues/8 a quick solution would be to mirror the baseline GCDs to `prod-exe`.

tianluyuan:
This requires a icetray 1.9.1 image to be built still.

closes #252


Dockerfile
```diff
@@ -1,10 +1,13 @@
...
 # Define the base image icetray version
 #
-ARG ICETRAY_VERSION=v1.7.2-ubuntu22.04-X64
+ARG ICETRAY_VERSION=v1.9.1-ubuntu22.04-X64
 
 FROM icecube/icetray:icetray-prod-$ICETRAY_VERSION as prod
 
+RUN mkdir -p /opt/i3-data/baseline_gcds && \
+    wget -nv -N -t 5 -P /opt/i3-data/baseline_gcds -r -l 1 -A *.i3* -nd http://prod-exe.icecube.wisc.edu/baseline_gcds/ && \
+    chmod -R u+rwX,go+rX,go-w /opt/i3-data/baseline_gcds
```
dsschult, line 10: Is it better to bind-mount these GCDs in, if we have CVMFS available?  Though, are these needed only on the client, or also on the server?

mlincett: > Is it better to bind-mount these GCDs in, if we have CVMFS available? Though, are these needed only on the client, or also on the server?

In #244 I was introducing the stating mechanisms for GCDs, then having them on CVFMS will be sufficient (falling back to http retrieval from `prod-exe`). But at the moment they are needed both on the server and the client. I hope I'll be able to finish that PR by January.

dsschult: Then I guess this is fine for now, since I believe the server uses this image without CVMFS.  (unless @ric-evans tells me I'm wrong)

ric-evans: the central server currently does not have access to CVMFS

dsschult:
Building the 1.9.1 icetray here: https://github.com/WIPACrepo/docker-icecube-icetray/actions/runs/7302075179/job/19900005995

Dockerfile
ric-evans: On one hand, I like putting the data in the image so that it is versioned. On the other hand, we may be carrying extra weight. @tianluyuan how much does this change increase the image size?

ric-evans (commented):
On one hand, I like putting the data in the image so that it is versioned. On the other hand, we may be carrying extra weight. @tianluyuan how much does this change increase the image size?

dsschult:
> On one hand, I like putting the data in the image so that it is versioned. On the other hand, we may be carrying extra weight. @tianluyuan how much does this change increase the image size?

It's about 600MB right now.

tianluyuan:
> > On one hand, I like putting the data in the image so that it is versioned. On the other hand, we may be carrying extra weight. @tianluyuan how much does this change increase the image size?
> 
> It's about 600MB right now.

Isn't it also already in the existing images now that are built on top of the icetray with the baseline GCDs?

ric-evans (approved):

dsschult:
> > > On one hand, I like putting the data in the image so that it is versioned. On the other hand, we may be carrying extra weight. @tianluyuan how much does this change increase the image size?
> > 
> > 
> > It's about 600MB right now.
> 
> Isn't it also already in the existing images now that are built on top of the icetray with the baseline GCDs?

I believe so.  I'm not actually too concerned when the image is on CVMFS, as it gets deduped.  It just takes longer to build, push, and pull the images, so CI always takes longer.  It's not a huge problem, but something to be aware of.

dsschult (approved):
Conditional on the tests passing, I'm good with this now.

PR #265, opened 2024-03-15
[minor] Pointed scans splinempe

G-Sommani:
Implemented for SplineMPE the pointed scans. Implemented a new reco class `SplineMPEPointed` which inherits from `SplineMPE` and performs a pointed scan around the `OnlineL2_SplineMPE` direction. It makes sense to use pointed scans with splinempe because usually the contours are very small and very near to the `OnlineL2_SplineMPE` direction. 
The precise feature in the `SplineMPEPointed` class that is making the scan pointed is the attribute `self.pointing_dir_names : List[str]`. If this is `None` (As it is by default in all recos), a full scan is performed. In `SplineMPEPointed` the attribite contains the list of all the possible names of the `OnlineL2_SplineMPE` direction, this tells `PixelsToReco` which direction to use for its attribute `self.pointing_ra_dec` and makes it different from `None`. Because `self.pointing_ra_dec` won't be `None`, the first pixels will be chosen by `choose_pixels_to_reconstruct_around_coord` in `pixels.py`. To implement new algorithms to do pointed scans it will just be necessary to create a new reco class with `self.pointing_dir_names` containing the desired `I3Particle` for the pointing.


tianluyuan:
Would it make sense to add a reco-algo in `skymap-scanner/recos`, call it e.g. `splinempe_online`, which inherits from the existing `class SplineMPE`? This subclass could then contain a flag with the keys for a targeted scan. The functionality in `start_scan.py` can check for such a flag, similar to existing flags like `self.reco.refine_time`, and be generic and applicable to other routines. This way we can also preserve the current splinempe full sky scans as available.

G-Sommani:
Hi all, I implemented the modification suggested by @tianluyuan. I created added a new reco-algo `SplineMPE_pointed` that inherits from `SplineMPE`. This reco has the flag `use_online_ra_dec` set to `True`.  All the other recos have it as well, but set to `False`. in `start_scan.py` the pixeler checks for this flag, and in case retrieves the onlinel2 direction.

tianluyuan (approved):
Thanks! this LGTM. @ric-evans?

skymap_scanner/recos/__init__.py
```diff
@@ -62,6 +62,12 @@ def get_vertex_variations() -> List[I3Position]:
         """Returns a list of vectors referenced to the origin that will be used to generate the vertex position variation."""
         pass
 
+    @staticmethod
+    @abstractmethod
+    def get_use_online_ra_dec() -> bool:
+        """Returns a boll whether to use online_ra_dec to point the scan."""
+        pass
+
```
ric-evans, line 70: Instead of defining this staticmethod in the base class (and setting `self.use_online_ra_dec = False` in most of the other derived classes), you could simplify this without any loss of info by detecting if `isint
SplineMPE_pointed`

skymap_scanner/recos/splinempe_pointed.py
```diff
@@ -0,0 +1,17 @@
...
+
+# mypy: ignore-errors
+
+from . import splinempe  
+from icecube import astro           
+from icecube.icetray import I3Frame
+from typing import Final, List
+from . import RecoInterface       
+
+class SplineMPE_pointed(splinempe.SplineMPE):
```
ric-evans, line 11: Can you use the case styling for the other reco classes, like `MillipedeWilks`?

skymap_scanner/recos/__init__.py
ric-evans, line 70: A dedicated method isn't necessary for returning this one variable, considering there isn't any additional logic. A publically accessible instance variable would suffice, and simplify the other classes

skymap_scanner/server/start_scan.py
```diff
@@ -126,6 +126,22 @@ def __init__(
...
+            particle_name_possibilities = ["OnlineL2_SplineMPE", "l2_online_SplineMPE"]
+            for particle_name in particle_name_possibilities:
+                if particle_name in p_frame.keys():
+                    online_dir = p_frame[particle_name].dir
+                    self.online_ra_dec = astro.dir_to_equa(
+                        online_dir.zenith,
+                        online_dir.azimuth,
+                        self.event_header.start_time.mod_julian_day_double
+                    )
+
```
ric-evans, line 144: Can you move this to a dedicated method in the reco class, and then return only the needed values? This will help keep these common functions brief.

skymap_scanner/recos/__init__.py
tianluyuan, line 1: I just noticed in the abstract base class you may also want to add `use_online_ra_dec: bool` at around L40, to ensure that it is defined in all derived classes.

ric-evans (changes requested):
Leaving a few code styling requests that should pay off in the long run

G-Sommani:
I think I implemented all the styling requests from @ric-evans . To retrieve `online_ra_dec`, I created the function `get_pointing_info` in `splinempe_pointed.py`. The check with `isinstance` is still performed in `start_scan.py`, to perform it in `recos` I would need to import there `splinempe_pointed`, which generates a circular import error.

G-Sommani:
I moved all the logic in `recos/__init__.py`. To do so, I created a publicly accessible instance variable `use_pointing`, which is `False` for all recos apart for `SplineMPE_pointed`. I haven't created any getter method for this instance variable.

skymap_scanner/recos/__init__.py
```diff
@@ -38,6 +39,7 @@ class RecoInterface(ABC):
     rotate_vertex: bool
     refine_time: bool
     add_fallback_position: bool
+    use_pointing: bool
```
ric-evans, line 42: I'm presuming that this is all `ABC` syntax magic :shrug: No change is needed if it works as designed, and I guess it does since it's an established pattern. Oh `ABC`...

skymap_scanner/server/start_scan.py
```diff
@@ -126,6 +126,8 @@ def __init__(
 
         self.omgeo = g_frame["I3Geometry"].omgeo
 
+        self.ang_dist, self.online_ra_dec = self.reco.get_pointing_info(p_frame)
```
ric-evans, line 129: These attributes appear innate (constant) to the reco algo instance. So duplicating these attrs is unnecessary and can eventually lead to some sticky situations. For example, `self.online_ra_dec` is changed here, but `self.reco`'s value doesn't. I think `self.reco.ang_dist` and `self.reco.online_ra_dec` would suffice

skymap_scanner/server/start_scan.py
```diff
@@ -163,7 +165,13 @@ def pixel_already_sent(pixel: Tuple[icetray.I3Int, icetray.I3Int]) -> bool:
...
         LOGGER.info(f"Looking for refinements for {nside_subprogression}...")
         #
-        pixels_to_refine = choose_pixels_to_reconstruct(self.nsides_dict, nside_subprogression)
+        pixels_to_refine = choose_pixels_to_reconstruct(
+            self.nsides_dict,
+            nside_subprogression,
+            ang_dist=self.ang_dist,
+            coord_ra_dec=self.online_ra_dec,
+        )
+        self.online_ra_dec = None
```
ric-evans, line 174: Is the intention that `self.online_ra_dec` is only used on the first invocation of this method, and ignored on subsequent calls? If that's the case, then a conditional around `nsides_dict` would be more robust, though that specific conditional logic may be better placed _in_ `choose_pixels_to_reconstruct()`, ie, always pass `coord_ra_dec` (or a renamed version of the var, if needed).

skymap_scanner/server/pixels.py
```diff
@@ -172,7 +172,7 @@ def choose_pixels_to_reconstruct(
     nsides_dict: NSidesDict,
     nside_progression: NSideProgression,
     ang_dist: float = 2.,
```
ric-evans, line 174: Do we rely on this default? @G-Sommani @tianluyuan @mlincett

skymap_scanner/recos/__init__.py
```diff
@@ -79,6 +81,25 @@ def traysegment(self, tray, name, logger, **kwargs: Any) -> None:
...
+            particle_name_possibilities = ["OnlineL2_SplineMPE", "l2_online_SplineMPE"]
+            for particle_name in particle_name_possibilities:
+                if particle_name in p_frame.keys():
+                    online_dir = p_frame[particle_name].dir
+                    online_ra_dec = astro.dir_to_equa(
+                        online_dir.zenith,
+                        online_dir.azimuth,
+                        p_frame["I3EventHeader"].start_time.mod_julian_day_double
+                    )
+        return ang_dist, online_ra_dec
```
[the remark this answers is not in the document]
ric-evans: For this function at large, is its logic specific to the pointed splinempe reco? (not a rhetorical question :smile:) If it is, then it should be put in the derived class--this would also allow you to remove `self.use_pointing`, since it'd rely on inheritance to get the right value(s) instead. The function in the base class would return the trivial values `(3.5, None)`, as it is today.

ric-evans (changes requested):
Since this PR proposes a few fundamental and important additions, I reviewed it especially intensely. We are close!

ric-evans:
Also, @G-Sommani, can you add `[minor]` to the PR title?

skymap_scanner/recos/__init__.py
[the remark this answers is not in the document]
G-Sommani: To reply to both comments, I imagine the pointing could be useful not just for SplineMPE, but even for other recos, even if I am not personally planning to implement it for the other recos. In case, different recos could prefer different values of `ang_dist`. For the pointing direction, `OnlineL2_SplineMPE` is what is usually sent out in the first online alert (the first GCN Notice), and I guess would work ad a good direction for pointing independently of the reconstruction algorithm that is after used. 
To make it brief, I believe the optimal solution would be:
1.  A pointing direction independent of the specific reconstruction algorithm (Now SplineMPE_pointed is the only reco algorithm with pointing but in future they could be more);
2.  A `ang_dist` variable which depends on the specific reconstruction.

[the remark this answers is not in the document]
G-Sommani: I implemented `ang_dist` as instance of `RecoInterface`. By default `ang_dist` is 3.5. If in future an algorithm needs a particular `ang_dist` this can be specified in the derived class. 

I renamed `online_ra_dec` to `pointing_ra_dec`, so it is more general and it doesn't implicitly say that the pointing direction is `OnlineL2_SplineMPE`. I kept `pointing_ra_dec` as an instance of `PixelsToReco`. Its value is now determined by the function `set_pointing_ra_dec` defined in `recos/__init__.py` but outside the class `RecoInterface`.

I added a the instance `pointing_dir_names` in `RecoInterface`, which by default is `None`. This instance contains the list of names for the `I3Particle` in the `p_frame` which contain the desired pointing direction. This instance is in `SplineMPE_pointed`, and if in future an algorithm will need a different direction this can be specified in the derived class.

The function `get_pointing_info` is now `set_pointing_dir`. It is defined in `recos/__init__.py` but it is not anymore a method of `RecoInterface`. I haven't moved it in `SplineMPE_pointed` because it can work as a general logic for any reco that will need the pointing. To implement the pointing it would just be necessary to define the name of the direction to use with `pointing_dir_names`.

skymap_scanner/server/start_scan.py
G-Sommani: I implemented a `len(nside_progression) == 1` in `choose_pixels_to_reconstruct()`, which is `True` only at the first invocation of the method.
It is important to use `choose_pixels_to_reconstruct_around_coord()` only at the first invocation for two reasons:
1. If it is always called, and the successive nsides the scanner won't be refining the pixels, but instead just always scan again the same area in the sky with always finer pixels (and for the nsides 512 and 1024 they become a huge number);
2. It may happen (very rarely, I would say once every ~200-300 events) that the minimum is around the border of the area selected for the scan, in this way the scanner is not forced to keep scanning always in the same area but can explore nearby as well.

skymap_scanner/server/pixels.py
G-Sommani: Maybe we can set here as well `ang_dist` to 3.5?

ric-evans: I'd rather the arg not have a default, then we'd rely on the reco_algo instance to supply it. That way we have a single source of truth

skymap_scanner/recos/splinempe_pointed.py
```diff
@@ -0,0 +1,16 @@
+"""IceTray segment for a pointed splinempe reco."""
+
+# mypy: ignore-errors
+ 
+from typing import Final
+
+from . import splinempe 
+from . import RecoInterface       
+
+class SplineMPE_pointed(splinempe.SplineMPE):
```
ric-evans, line 10: You can call this `SplineMPEPointed` and existing logic should transform the casing from the CL arg

ric-evans (changes requested):
It's looking great. The only thing I noticed is the class naming, to match python conventions. Otherwise, the comment on the default value will need others' input, and may be best for a separate PR

skymap_scanner/server/pixels.py
```diff
@@ -182,13 +182,13 @@ def choose_pixels_to_reconstruct(
 
     Some of the pixel returned may have previously been generated.
     """
-    # special case if we have MC truth
-    if mc_ra_dec:
-        LOGGER.debug("Getting pixels around MC truth...")
-        # scan only at max_nside around the true minimum
-        return choose_pixels_to_reconstruct_around_MCtruth(
+    # special case if we have a given coordinate, select pixels only at first nside
+    if coord_ra_dec and len(nside_progression) == 1 :
```
tianluyuan, line 186: What happens if `coord_ra_dec` is set but `len(nside_progression) !=1`? Maybe this should be an `assert` instead so that it fails if the user passes multiple sides?

ric-evans: If that's the assumption, then let's (also) put that validation check at the very top of the pipeline where the CL args are retrieved

G-Sommani: The function `choose_pixels_to_reconstruct_around_coord` must run only when the pixels for the first `nside` are chosen, and not when pixels are selected for refinements. When the scanner chooses the pixels for higher nsides, then `coord_ra_dec` is set but  `len(nside_progression) !=1 (here the `nside_progression` is what in `start_scan.py` is the `nside_subprogression`, and at the first iteration is just a `dict` with `len(nside_subprogression) = 1`), and this is necessary as we discussed here: https://github.com/icecube/skymap_scanner/pull/265#discussion_r1553791298.

G-Sommani: > If that's the assumption, then let's (also) put that validation check at the very top of the pipeline where the CL args are retrieved

`start_scan.py` calls `choose_pixels_to_reconstruct()` sending the `nside_subprogression`, which just tells until which `nside` the pixels must be selected. The `nside_progression` that is sent as an input to `start_scan` is the complete progression and is not what is checked here, so I would avoid a validation check at the beginning of the pipeline, as the `nside_progression` that is sent as an arg is not the `nside_progression` that is checked here.

G-Sommani:
I changed the name of the class to `SplineMPEPointed`, as requested by @ric-evans .

skymap_scanner/server/pixels.py
ric-evans: I believe I understand--in other words, `len(nside_subprogression)` may be > 1 for scans that define `coord_ra_dec` (aka `splinempe_pointed`). Is that correct?

ric-evans (approved):
Great work! I think the PR description may need updating to reflect your new changes, but all in all, it is great to see this previously buried functionality revived

skymap_scanner/server/pixels.py
G-Sommani: Yes, at all the pixel refinements `len(nside_subprogression)` will be `>1` and the pixels won't be chosen by `choose_pixels_to_reconstruct_around_coord()`

PR #272, opened 2024-09-20
Pin iminuit

tianluyuan:
Pin iminuit<2.27.0. 2.27.0 introduced a change in the minimization algorithm (scikit-hep/iminuit#1009) that does not ensure reproducibility and led to the tests failing.


tianluyuan:
See also scikit-hep/iminuit#1041

PR #275, opened 2024-10-01
Implement Retries and Timeout for File Download

ric-evans:
Previously, files were retrieved with `subprocess` + `wget`. Now, it uses `requests` with a fixed timeout and retries, plus exponential backoff.

### DIFF
--- a/skymap_scanner/utils/data_handling.py
+++ b/skymap_scanner/utils/data_handling.py
@@ -1,9 +1,11 @@
 """data_handling.py."""
 
 import logging
-import subprocess
+import time
 from pathlib import Path
 from typing import List
+
+import requests
 
 from .. import config as cfg
 
@@ -41,12 +43,12 @@ class DataStager:
                     LOGGER.debug("File is available on staging path.")
                 else:
                     LOGGER.debug("Staging from HTTP source.")
-                    self.fetch_file(basename)
+                    self.download_file(basename)
 
             else:
                 LOGGER.debug(f"File {basename} is available at {filepath}.")
 
-    def fetch_file(self, basename: str):
+    def download_file(self, basename: str):
         """Retrieves a file from the HTTP source.
 
         Args:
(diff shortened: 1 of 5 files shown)


skymap_scanner/utils/data_handling.py
```diff
@@ -55,24 +57,50 @@ def fetch_file(self, basename: str):
...
+                    raise RuntimeError(
+                        f"Download failed after {cfg.REMOTE_DATA_DOWNLOAD_RETRIES} retries: {e}"
+                    ) from e
+
+        # Step 2: Write the file
+        try:
+            with open(dest, "wb") as file:
+                for chunk in response.iter_content(chunk_size=8192):
+                    file.write(chunk)
+        except IOError as e:
```
tianluyuan, line 95: Why does this need to be caught in a try except? Isn't it clearer to let it raise as an IOError?

skymap_scanner/utils/data_handling.py
```diff
@@ -55,24 +57,50 @@ def fetch_file(self, basename: str):
...
+
+        def backoff_sleep(attempt: int):
+            """Sleep with exponential backoff."""
+            sleep_duration = 2**attempt  # Exponential backoff: 2, 4, 8 seconds...
+            LOGGER.info(f"Retrying file download in {sleep_duration} seconds...")
+            time.sleep(sleep_duration)
+
+        # Step 1: Download the file
+        attempt = 0
+        while True:
```
tianluyuan, line 71: Why not use a for loop?

skymap_scanner/utils/data_handling.py
ric-evans: The idea is to tell the user why there's an error. Since it's using `from e`, we still get to see the original error.

skymap_scanner/utils/data_handling.py
ric-evans: Using `itertools.count()` now

skymap_scanner/utils/data_handling.py
```diff
@@ -55,24 +58,48 @@ def fetch_file(self, basename: str):
...
+            except requests.exceptions.RequestException as e:
+                if attempt > cfg.REMOTE_DATA_DOWNLOAD_RETRIES:  # 'attempt' is 1-indexed
+                    raise RuntimeError(
+                        f"Download failed after {cfg.REMOTE_DATA_DOWNLOAD_RETRIES} retries: {e}"
+                    ) from e
+
+        # Step 2: Write the file
+        try:
+            with open(dest, "wb") as file:
+                for chunk in response.iter_content(chunk_size=8192):
```
tianluyuan, line 92: Anything special about 8192?

tianluyuan (approved):

skymap_scanner/utils/data_handling.py
ric-evans: Nothing special, it just limits how much is read into memory. I don't know how large our downloads are.

PR #278, opened 2025-02-24
Fix CVMFS Image Removal GH Action

### LINKED ISSUE #263: Fix CVMFS Singularity-Delete GHA
Images aren't being marked for removal when branches are deleted: https://github.com/WIPACrepo/cvmfs-actions/blob/main/docker_images.txt

ric-evans:
closes #263


ric-evans:
test what happens when branch is deleted

ric-evans:
will need to merge so the workflow on main will be updated

Issue #242, opened 2023-11-29
Larger Relative Tolerance Variation in `splinempe`


ric-evans:
In [test-run-realistic (splinempe, run00136662-evt000035405932-BRONZE.pkl)](https://github.com/icecube/skymap_scanner/actions/runs/7036374924/job/19148894289#logs), the measured variation is seen at `0.1689` for a single pixel. This is higher than the accepted `0.1`.

```
[
         [
            1,
            1261.4054977969026,
            NaN,
            NaN
         ],
         [
            1,
            1079.0788843076734,
            NaN,
            NaN
         ],
         [
            0.0,
            0.16896504614322816,
            NaN,
            NaN
         ],
         [
            true,
            false,
            true,
            true
         ]
      ],
...
```

tianluyuan:
It looks like the XYZT reconstructions are even more susceptible to this issue with reproducibility. Copying over results from one CI test still led to failures on the subsequent run. The CI jobs were run on the same machine so it's possible there's some randomness in the seed chain that's causing this.

ric-evans:
> It looks like the XYZT reconstructions are even more susceptible to this issue with reproducibility. Copying over results from one CI test still led to failures on the subsequent run. The CI jobs were run on the same machine so it's possible there's some randomness in the seed chain that's causing this.

as seen in https://github.com/icecube/skymap_scanner/pull/281

PR #281, opened 2025-03-12
[minor] Write reconstructed position and time

JannisNe:
Add x, y and z position from the reco `I3Position` and the corresponding time to the result file for each scanned pixel. This allows to display the reconstructed track later for visualization. There is a [corresponding PR in `skyreader`](https://github.com/icecube/skyreader/pull/47).


tianluyuan:
@JannisNe I created a container image from this branch. Once that's up on cvmfs you should be able to run it via something like `skymist scan event --docker_tag feature-write-position-and-time --additional_tag xyzt.test 138449 20481611 millipede_wilks fine 500` 

@ric-evans do you know what's causing the ewms installation failures? Those seem to be the tests still failing.

tianluyuan:
Just to follow up, I ran a test with the docker image and the new format seems to be working. This is the result for scan ID `e726b00989de4923a2e887048ab66891`

```
{
    "nside-1": {
        "columns": [
            "index",
            "llh",
            "E_in",
            "E_tot",
            "X",
            "Y",
            "Z",
            "T"
        ],
        "data": [
            [
                0,
                1416.9539526405474,
                15619.416470487953,
                15328321.963551521,
                115.5865566877651,
                397.1009857956931,
                -211.40070101741426,
                9763.02271870064
            ],
            [
                1,
                1369.522861787743,
                0.0,
                2253461.1160950465,
                57.54951154635373,
                530.9337137879596,
                -290.0320568844377,
                9933.696969297525
            ],
...
```

setup.cfg
```diff
@@ -23,7 +23,7 @@ branch = main
 install_requires =
 	ewms-pilot==0.19.0 # the extras here are the same as 'oms-mqclient', so just define one place
 	healpy
-	icecube-skyreader
+	icecube-skyreader>=1.4.2
```
ric-evans, line 26: This doesn't need to be pinned. Any fresh build of the project will always pull the latest.

ric-evans (changes requested):
Question: Is there any reason not to test for the correct position and time values?

The scanner is using `version=1` (default value) of the `EventMetadata` instance (https://github.com/icecube/skymap_scanner/blob/main/skymap_scanner/utils/extract_json_message.py#L190-L196), but I don't see these values being accessed. This requires running a few branch/test scans on skydriver, manually (by helper script) confirming the results, then updating this repo's test files.

setup.cfg
tianluyuan: That's true, however I guess it wouldn't hurt to be explicit here? Also I'm not sure what would happen in the case where an older version of the dependency is already installed and this isn't specified.

tianluyuan:
> Question: Is there any reason not to test for the correct position and time values?
> 
> The scanner is using `version=1` (default value) of the `EventMetadata` instance (https://github.com/icecube/skymap_scanner/blob/main/skymap_scanner/utils/extract_json_message.py#L190-L196), but I don't see these values being accessed. This requires running a few branch/test scans on skydriver, manually (by helper script) confirming the results, then updating this repo's test files.

Yes the version-dependent checks are defined in skyreader. Specifically for the `isclose` method, it only checks fields that exist in the intersection of both results. If the tests files are updated then we could tighten the tolerances on the XYZT and it should check all those fields as well. Do you have the scripts to do those updates? I didn't do that as I didn't want to manually update it one-by-one.

ric-evans:
> > Question: Is there any reason not to test for the correct position and time values?
> > The scanner is using `version=1` (default value) of the `EventMetadata` instance ([`main`/skymap_scanner/utils/extract_json_message.py#L190-L196](https://github.com/icecube/skymap_scanner/blob/main/skymap_scanner/utils/extract_json_message.py?rgh-link-date=2025-03-17T16%3A47%3A27Z#L190-L196)), but I don't see these values being accessed. This requires running a few branch/test scans on skydriver, manually (by helper script) confirming the results, then updating this repo's test files.
> 
> Yes the version-dependent checks are defined in skyreader. Specifically for the `isclose` method, it only checks fields that exist in the intersection of both results. If the tests files are updated then we could tighten the tolerances on the XYZT and it should check all those fields as well. Do you have the scripts to do those updates? I didn't do that as I didn't want to manually update it one-by-one.

These test files would need updating: https://github.com/icecube/skymap_scanner/tree/feature-write-position-and-time/tests/data/results_json. Keepng the `reco_pixel_single/` files unchanged is fine. I have a script I can use for running those events. I'll generate those and follow up with new data.

tianluyuan:
I updated the tolerances and reverted the try-except. Tests should hopefully pass once the test files are updated @ric-evans

ric-evans:
> @ric-evans looks like we're running into possibly #242 for the test_realistic SplineMPE scans. Basically results are not reproducible exactly.
> 
> Would it make sense to go back to testing against the version 0 files for those three only?

Did the rtol change for splinempe with version 1? I'd like to see the new field values be tested for all reco algos.

tianluyuan:
> > @ric-evans looks like we're running into possibly #242 for the test_realistic SplineMPE scans. Basically results are not reproducible exactly.
> > Would it make sense to go back to testing against the version 0 files for those three only?
> 
> Did the rtol change for splinempe with version 1? I'd like to see the new field values be tested for all reco algos.

The tolerances for existing fields did not change, but the XYZT fields are new and set to some reasonable values. If possible, we should try not to increase those tolerances just to get the SplineMPE tests to pass. Alternatively, it could also be fine to let them fail in the hopes that there's some solution eventually :)

ric-evans (approved):
If the tests pass, all is good

PR #282, opened 2025-04-04
Realtime version passthrough

tianluyuan:
Pass through the realtime version string to the `RecoInterface` object in order to give reconstructions the ability to determine what keys to use.

For example, for pointed scans, the particle around which to scan used to be a list of particles names that contained variations across all realtime versions. This is brittle and prone to unexpected behavior. By passing the version string to `RecoInterface` instances the appropriate key can be chosen based on that.

Additionally, remove `INPUT_PULSES_NAME` in config.py in favor of accessing it via a static method in the RecoInterface via the `INPUT_KEYS_MAP` defined in config.py. This removes the need to copy / rename pulseseries in the frame, and also allows for future key name changes with the realtime format.

Changes a few base class variables that were later shadowed by instance variables into instance variables in the base class itself.


skymap_scanner/recos/__init__.py
```diff
@@ -35,20 +35,21 @@ class RecoInterface(ABC):
...
-    add_fallback_position: bool
 
     # List of spline filenames required by the class.
     # The spline files will be looked up in pre-defined local paths or fetched from a remote data store.
     SPLINE_REQUIREMENTS: List[str] = list()
 
     @abstractmethod
-    def __init__(self):
-        pass
+    def __init__(self, realtime_format_version):
```
ric-evans, line 44: The new arg needs a typehint

ric-evans: Also, do you want to store `realtime_format_version` in the instance?

skymap_scanner/recos/dummy.py
```diff
@@ -29,7 +29,8 @@
 class Dummy(RecoInterface):
     """Logic for a dummy reco."""
 
-    def __init__(self):
+    def __init__(self, realtime_format_version):
```
ric-evans, line 32: typehint (since the rest of the class uses them)

skymap_scanner/utils/extract_json_message.py
```diff
@@ -51,29 +51,19 @@ def extract_GCD_diff_base_filename(frame_packet):
...
     cache_dir: str,
     GCD_dir: str
-) -> Tuple[EventMetadata, dict]:
+) -> Tuple[EventMetadata, dict, str]:
 
     _validate_cache_dir(cache_dir=cache_dir)
-    pulses_name = _get_pulses_name(event_dict=event_dict)
+    # Some JSON events may not have the 'version' attribute.
+    # In such case we default to "no-version".
+    realtime_format_version = event_dict["value"].get("version", "no-version")
```
ric-evans, line 66: Would `None` or `""` suffice?

skymap_scanner/server/start_scan.py
```diff
@@ -328,6 +332,7 @@ async def scan(
     from_clients_queue: mq.Queue,
     nside_progression: NSideProgression,
     predictive_scanning_threshold: float,
+    realtime_format_version: str
```
ric-evans, line 335: needs a trailing comma--this decreases diffs if a future change adds another arg

skymap_scanner/server/start_scan.py
```diff
@@ -347,6 +352,7 @@ async def scan(
         output_particle_name=cfg.OUTPUT_PARTICLE_NAME,
         reco_algo=reco_algo,
         event_metadata=event_metadata,
+        realtime_format_version=realtime_format_version
```
ric-evans, line 355: See comment on trailing comma--there are a few other cases where this applies but I won't add comments

ric-evans (changes requested):
A few comments, a couple of questions, nothing big

skymap_scanner/recos/splinempe.py
```diff
@@ -238,23 +238,23 @@ def log_frame(frame):
...
+                logger.debug(f"Seed from {self.vertex_seed_source}:")
                 logger.debug(frame[cfg.INPUT_POS_NAME])
                 # logger.debug(f"Seed from OnlineL2_SplineMPE:")
                 # logger.debug(frame["OnlineL2_SplineMPE"].pos)
                 # logger.debug(frame["OnlineL2_SplineMPE"].time)
 
             tray.Add(notify_seed)
 
-        elif cls.vertex_seed_source == "OnlineL2_SplineMPE":
+        elif self.vertex_seed_source == "OnlineL2_SplineMPE":
```
G-Sommani, line 257:
```suggestion
        elif self.vertex_seed_source == self.pointing_dir_name = config.INPUT_KEY_NAMES_MAP.get(
                realtime_format_version,
                config.DEFAULT_INPUT_KEY_NAMES).l2_splinempe :
```

skymap_scanner/recos/splinempe.py
```diff
@@ -238,23 +238,23 @@ def log_frame(frame):
...
             # First vertex seed is extracted from OnlineL2 reco.
             def extract_seed(frame):
                 seed_source = "OnlineL2_SplineMPE"
```
G-Sommani, line 260:
```suggestion
                seed_source = self.pointing_dir_name = config.INPUT_KEY_NAMES_MAP.get(
                    realtime_format_version,
                    config.DEFAULT_INPUT_KEY_NAMES).l2_splinempe:
```

skymap_scanner/recos/splinempe.py
tianluyuan: I've updated it to take save the l2_splinempe name in the splinempe instance

skymap_scanner/config.py
```diff
@@ -74,10 +75,10 @@
 STATEDICT_GCDQP_PACKET: Final = "GCDQp_packet"
 STATEDICT_BASELINE_GCD_FILE: Final = "baseline_GCD_file"
 STATEDICT_NSIDES: Final = "nsides"
-STATEDICT_INPUT_PULSES: Final = "input_pulses_name"
 #
 MSG_KEY_RECO_ALGO: Final = "reco_algo"
 MSG_KEY_PFRAME_PKL_B64: Final = "pframe_pkl_b64"
+MSG_KEY_FORMAT_VERSION: Final = "format_version"
```
ric-evans, line 81: I recommend renaming this key `realtime_format_version` and the variable `MSG_KEY_REALTIME_FORMAT_VERSION` to match the other use cases. This way, if we add another "version" variable in the future, there'd be no confusion (or refactoring needed).

skymap_scanner/config.py
```diff
@@ -53,14 +55,13 @@
...
+KeyNames = namedtuple('KeyNames', 'pulseseries l2_splinempe')
+INPUT_KEY_NAMES_MAP: Final[dict[str, KeyNames]] = {
+    "2021a": KeyNames("SplitUncleanedInIcePulses", "OnlineL2_SplineMPE"),
+    "2023a": KeyNames("SplitInIcePulses", "l2_online_SplineMPE"),
+    "2024a": KeyNames("SplitInIcePulses", "l2_online_SplineMPE"),
 }
-DEFAULT_INPUT_PULSES_NAME: Final = "SplitUncleanedInIcePulses"
-
-INPUT_PULSES_NAME = "SplitUncleanedInIcePulses"
+DEFAULT_INPUT_KEY_NAMES: Final = KeyNames("SplitUncleanedInIcePulses", "OnlineL2_SplineMPE")
```
ric-evans, line 64: These are keyed by the `realtime_format_version`, right? Can you make it clear that they are, either in a comment or by renaming the variable(s)?

ric-evans (changes requested):
Two more things

skymap_scanner/config.py
tianluyuan: Added a comment

skymap_scanner/config.py
ric-evans: I think you missed the value--I suggest renaming the key in the message objects/ json files.

ric-evans (approved):

PR #285, opened 2025-07-30
Fix apptainer dummy tests

ric-evans:
Follow up from https://github.com/icecube/skymap_scanner/pull/284


ric-evans:
merged back into #284

PR #311, opened 2025-11-13
Update SplineMPE tests

tianluyuan:
Remove the --check-different-versions-ok flag


ric-evans (changes requested):
running into the problem described in https://github.com/icecube/skymap_scanner/issues/242#issuecomment-2733860943

tianluyuan:
Seems like positional coordinates are triggering failures even after updating the references..

